PR #23, opened 2019-05-20
Mqtt cleanup and better handle requests

GwendalRaoul:
This pull request contains two things:
- First two commits are to better isolate the Mqtt client to the transport itself
- Last one is to handle requests on dedicated threads

### DIFF
--- a/python_transport/wirepas_gateway/protocol/mqtt_wrapper.py
+++ b/python_transport/wirepas_gateway/protocol/mqtt_wrapper.py
@@ -1,0 +1,217 @@
+# Wirepas Oy licensed under Apache License, Version 2.0
+#
+# See file LICENSE for full license details.
+#
+import queue
+import socket
+import ssl
+from select import select
+from threading import Thread, current_thread
+from time import sleep
+
+from paho.mqtt import client as mqtt
+
+
+class MQTTWrapper(Thread):
+    """
+    Class to manage the MQTT main thread and be able to share it with other services
+    In this case, it allows to have all the related mqtt activity happening on same thread
+    to avoid any dead lock from mqtt client.
+    """
+
+    def __init__(self, logger, username, password, host, port, secure_auth=True, tlsfile=None,
+                 on_termination_cb=None, on_connect_cb=None):
+        Thread.__init__(self)
+        self.daemon = True
+        self.running = False
+        self.logger = logger
+        self.on_termination_cb = on_termination_cb
+        self.on_connect_cb = on_connect_cb
+
+        self._client = mqtt.Client()
+        if secure_auth:
+            try:
+                self._client.tls_set(
+                    tlsfile,
+                    certfile=None,
+                    keyfile=None,
+                    cert_reqs=ssl.CERT_REQUIRED,
+                    tls_version=ssl.PROTOCOL_TLSv1_2,
+                    ciphers=None,
+                )
+            except:
+                self.logger.error(
+                    "Cannot use secure authentication. attempting unsecure connection"
+                )
+
+        self._client.username_pw_set(username, password)
+        self._client.on_connect = self._on_connect
+
+        try:
+            self._client.connect(host, port, keepalive=60)
+        except (socket.gaierror, ValueError) as e:
+            self.logger.error("Cannot connect to mqtt {}".format(e))
+            exit(-1)
+
+        # Set options to initial socket
+        self._client.socket().setsockopt(socket.SOL_SOCKET, socket.SO_SNDBUF, 2048)
+
+        self._publish_queue = SelectableQueue()
+
+    def _on_connect(self, client, userdata, flags, rc):
+        if rc != 0:
+            self.logger.error("MQTT cannot connect {}".format(rc))
+            return
+
+        if self.on_connect_cb is not None:
+            self.on_connect_cb()
+
+    def _do_select(self, sock):
+        # Select with a timeout of 1 sec to call loop misc from time to time
+        r, w, e = select(
+            [sock, self._publish_queue],
+            [sock] if self._client.want_write() else [],
+            [],
+            1,
+        )
+
+        # Check if we have something to publish
+        if self._publish_queue in r:
+            try:
+                # Publish everything. Loop is not necessary as
+                # next select will exit immediately if queue not empty
+                while True:
+                    topic, payload, qos, retain = self._publish_queue.get()
+                    self._client.publish(topic, payload, qos=qos, retain=retain)
+
+                    # FIX: read internal sockpairR as it is written but
+                    # never read as we don't use the internal paho loop
+                    # but we have spurious timeout / broken pipe from
+                    # this socket pair
+                    try:
+                        self._client._sockpairR.recv(1)
+                    except Exception:
+                        # This socket is not used at all, so if something is wrong,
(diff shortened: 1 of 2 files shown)


python_transport/wirepas_gateway/protocol/mqtt_wrapper.py
```diff
@@ -0,0 +1,217 @@
...
+
+
+class MQTTWrapper(Thread):
+    """
+    Class to manage the MQTT main thread and be able to share it with other services
+    In this case, it allows to have all the related mqtt activity happening on same thread
+    to avoid any dead lock from mqtt client.
+    """
+
+    def __init__(self, logger, username, password, host, port, secure_auth=True, tlsfile=None,
```
PFigs: The tlsfile variable should be renamed to ca_certs, since it is what it is. 

It will also avoid confusion when passing in the client certificates, which should be exposed with this change. 

Here's what I think should be added to the constructor:

-   certfile=None
-   keyfile=None
-   cert_reqs=ssl.CERT_REQUIRED
-   tls_version=ssl.PROTOCOL_TLS
-   ciphers=None

We also don't seem to have a way of propagating a signal to exit from the main loop.

python_transport/wirepas_gateway/protocol/mqtt_wrapper.py
```diff
@@ -0,0 +1,217 @@
...
+            except:
+                self.logger.error(
+                    "Cannot use secure authentication. attempting unsecure connection"
+                )
+
+        self._client.username_pw_set(username, password)
+        self._client.on_connect = self._on_connect
+
+        try:
+            self._client.connect(host, port, keepalive=60)
```
PFigs: use a class variable to control the keep alive

python_transport/wirepas_gateway/protocol/mqtt_wrapper.py
```diff
@@ -0,0 +1,217 @@
...
+                        # This socket is not used at all, so if something is wrong,
+                        # not a big issue. Just keep going
+                        pass
+
+            except TimeoutError:
+                self.logger.debug("Timeout to send payload: {}".format(payload))
+                # In theory, mqtt client shouldn't loose the last packet
+                # If it is not the case, following line could be uncommented
+                # self._publish_queue.put((topic, payload, qos, retain))
+                raise
```
PFigs: It should mark a flag which triggers a clean exit

python_transport/wirepas_gateway/protocol/mqtt_wrapper.py
```diff
@@ -0,0 +1,217 @@
...
+        self.logger.error("MQTT, unexpected disconnection")
+
+        # Socket is not opened anymore, try to reconnect
+        while True:
+            try:
+                self._client.reconnect()
+                break
+            except Exception:
+                # Retry to connect in 1 sec
+                sleep(1)
```
PFigs: Have you considered adding a counter and exiting after a few attempts?

python_transport/wirepas_gateway/protocol/mqtt_wrapper.py
```diff
@@ -0,0 +1,217 @@
...
+        # Set options to new reopened socket
+        self._client.socket().setsockopt(socket.SOL_SOCKET, socket.SO_SNDBUF, 2048)
+        return self._client.socket()
+
+    def set_last_will(self, topic, data):
+        # Set Last wil message
+        self._client.will_set(topic, data, qos=2, retain=True)
+
+    def run(self):
+        while True:
```
PFigs: We should use a flag that allows us to exit cleanly from the loop.

python_transport/wirepas_gateway/protocol/mqtt_wrapper.py
```diff
@@ -0,0 +1,217 @@
...
+
+                self._do_select(sock)
+            except TimeoutError:
+                self.logger.error("Timeout in connection, force a reconnect")
+                self._client.reconnect()
+            except Exception:
+                # If an exception is not catched before this point
+                # All the transport module must be stopped in order to be fully
+                # restarted by the managing agent
+                self.logger.exception("Unexpected exception in MQTT wrapper Thread")
```
PFigs: I would prefer to see the exception breaking and on the outside of the loop you could add the cleanup code.

python_transport/wirepas_gateway/transport_service.py
```diff
@@ -233,6 +247,7 @@ def on_sink_disconnected(self, name):
         self.logger.info("Sink disconnected, sending new configs")
         self._send_asynchronous_get_configs_response()
 
+    @deferred_thread
```
PFigs: Could we run into a race condition when calling a few resources used by these threads? For example, the sink_manager.get_sink?

python_transport/wirepas_gateway/protocol/mqtt_wrapper.py
GwendalRaoul: Yes agree.
In this change I'll just update the tlsfile->ca_certs name.
Adding new parameter will be addressed in a different cr

python_transport/wirepas_gateway/protocol/mqtt_wrapper.py
GwendalRaoul: Done

python_transport/wirepas_gateway/protocol/mqtt_wrapper.py
GwendalRaoul: This part of code is not needed so I'll remove it.
This exception is catched in the main Thread of this class and retry to reconnect.

python_transport/wirepas_gateway/protocol/mqtt_wrapper.py
GwendalRaoul: Yes good idea. But exiting here mean exiting the full transport module. It is what we want?

python_transport/wirepas_gateway/protocol/mqtt_wrapper.py
GwendalRaoul: Done

python_transport/wirepas_gateway/protocol/mqtt_wrapper.py
GwendalRaoul: Done

python_transport/wirepas_gateway/transport_service.py
GwendalRaoul: No it shouldn't. The only shared ressource being the dbus object. This is the only place where the different Threads will release the GIL (IO blocking). So Threads will get the CPU only when one is waiting on dbus and this part should be thread safe.
I did a test by creating long delay on sink service to serve the request and doing multiple time the same request on MQTT. Threads are all blocked on dbus and get access to it one by one.
Of course their is still a possibility of sink service disapearing but it is exactly the same with current implementation and already handled.

python_transport/wirepas_gateway/protocol/mqtt_wrapper.py
PFigs: yes

python_transport/wirepas_gateway/transport_service.py
PFigs: good with me!

PR #42, opened 2019-06-18
Docker images building procedure

### LINKED ISSUE #15: Configuration from template file
*Problem"
Currently the docker builds rely on an entry point to evaluate environment parameters that are meant for the transport or sink service. The entry point becomes rather complex and difficult to extend and mantain.

*Proposal*
Generate a settings.yaml/settings.ini file, based on the environment parameters. The template file will contain the default values and in case they are propagated, the final file will have the corresponding parameters.

PFigs:
This commit simplifies the building of docker images by
relying on a docker-comopse for each type of image.

It is still possible to reduce the build to a single
compose file + a shell script to set the correct
tags and paths. However this is left for future
consideration.

Closes #15 and #26


PFigs:
Travis update to follow

Issue #28, opened 2019-06-12
Expose MQTT client properties


PFigs:
Expose to the transport service arguments to control the MQTT session. 

These include:
* Client certificates 
* SSL options
* Session cleanup (use client id as gateway id)

https://github.com/wirepas/gateway/blob/master/python_transport/wirepas_gateway/utils/argument_tools.py

https://github.com/wirepas/gateway/blob/df58d1c2c64c5860e431d64690797a2ab1003844/python_transport/wirepas_gateway/transport_service.py#L591

PR #49, opened 2019-06-21
Transport: expose all mqtt standard parameters (#28)

GwendalRaoul:
- Allign them with backend-client
- Keep old ones with deprecated mention
- Directly send global settings in different classes to avoid specifying
  all the different parameters everywhere

### DIFF
--- a/python_transport/wirepas_gateway/protocol/mqtt_wrapper.py
+++ b/python_transport/wirepas_gateway/protocol/mqtt_wrapper.py
@@ -26,8 +26,11 @@ class MQTTWrapper(Thread):
     # Reconnect timeout in Seconds
     TIMEOUT_RECONNECT_S = 120
 
-    def __init__(self, logger, username, password, host, port, secure_auth=True, ca_certs=None,
-                 on_termination_cb=None, on_connect_cb=None):
+    def __init__(self,
+                 settings,
+                 logger,
+                 on_termination_cb=None,
+                 on_connect_cb=None):
         Thread.__init__(self)
         self.daemon = True
         self.running = False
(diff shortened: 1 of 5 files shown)


python_transport/wirepas_gateway/protocol/mqtt_wrapper.py
```diff
@@ -26,36 +26,41 @@ class MQTTWrapper(Thread):
...
-                    ciphers=None,
+                    ca_certs=settings.mqtt_ca_certs,
+                    certfile=settings.mqtt_certfile,
+                    keyfile=settings.mqtt_keyfile,
+                    cert_reqs=settings.mqtt_cert_reqs,
+                    tls_version=settings.mqtt_tls_version,
+                    ciphers=settings.mqtt_ciphers
                 )
-            except:
+            except Exception as e:
```
PFigs: Should we actually force an exit instead of attempting an unsecured connection?

python_transport/wirepas_gateway/utils/argument_tools.py
```diff
@@ -122,75 +122,255 @@ def add_file_settings(self):
...
+
+    def add_deprecated_args(self):
+        """ Deprecated mqtt arguments in order to keep backward compatibility """
+        self.deprecated.add_argument(
+            "-s",
+            "--host",
+            default=None,
+            type=str,
+            help=(
+                "Deprecated, pleas use -mqtt_hostname")
```
PFigs: I would suggest we state that this will be dropped from version 2.x onwards

python_transport/wirepas_gateway/protocol/mqtt_wrapper.py
GwendalRaoul: Yes, you're right!

python_transport/wirepas_gateway/utils/argument_tools.py
GwendalRaoul: Yes, I'll update the comment

PR #60, opened 2019-07-12
Single Dockerfile for multi-arch builds

PFigs:
This commit removes the need of keeping two Dockerfiles for
different architectures.

However, it still relies on docker-compose to build release
versions or the .ci/build-images.sh for continuous integration
builds.

In terms of release builds, these will only be built when a tag
is given to ci/travis. This is in preparation for the multi-arch
push to docker hub.

There is still a need in the future to accomodate other OSes inside
the container folder, such as Alpine.


.ci/build-images.sh
```diff
@@ -61,9 +82,23 @@ function _pull_dev_dependencies
...
 {
-    _build "arm"
-    _build "x86"
-    _build "dev"
+
+    # builds x86 and arm images based on manifest files
+    if [[ ! -z ${BUILD_TAG} ]]
+    then
+        GIT_MANIFEST_BRANCH=refs/tags/${BUILD_TAG}
+        _build "${DOCKERFILE_PATH}/arm" "arm" "--no-cache"
```
GwendalRaoul: I don't see the point of these two lines compare to the two ones after the end of if

PFigs: I forgot to remove the ones outside the if case

PR #69, opened 2019-07-15
Run yaml parameters through argparse

### LINKED ISSUE #67: Settings from yaml file do not respect variable type
When evaluating parameters from a yaml file the type is not respected.

One possible way to fix this is to pass the arguments through argparse.

PFigs:
The type of input parameters is not respected when reading them
from a yaml file. This happens due to the fact that the dictionary
obtained from yaml is used to overwrite entries in the argument
list.

This change calls the parse_known_args from the argparse to avoid
issues with the value type.

Closes #67


python_transport/wirepas_gateway/utils/argument_tools.py
```diff
@@ -100,40 +95,58 @@ def unkown_arguments(self):
...
+                    arglist.append(arg)
+
+            # Override self._arguments as there are parameters from file
             self._arguments, self._unknown_arguments = self.parser.parse_known_args(
                 arglist
             )
 
+        except FileNotFoundError as e:
+            # Cannot find the setting file
+            if self._arguments.settings == "settings.yml":
```
PFigs: Should we in fact default the file to None?

GwendalRaoul: Yes maybe the best solution. It breaks the backward compatibility but not sure it was really used with default value

Issue #77, opened 2019-07-16
Change the debug variable name to make it consistent with others


PFigs:
Currently the transport service uses DEBUG_LEVEL to acquire the transport service level. However, most other environment parameters are preceded by WM_. 

https://github.com/wirepas/gateway/blob/41ca5e966806c8e2122f5c9956ca20d0790b2a97/python_transport/wirepas_gateway/transport_service.py#L611

PR #80, opened 2019-07-17
Update transport env

PFigs:
Targets #77 and enhances logging message

### DIFF
--- a/python_transport/wirepas_gateway/transport_service.py
+++ b/python_transport/wirepas_gateway/transport_service.py
@@ -613,7 +613,7 @@ def main():
 
     """
     ParserHelper()
-    parse = ParserHelper(description="Default arguments")
+    parse = ParserHelper(description="Wirepas Gateway Transport service arguments")
 
     parse.add_file_settings()
     parse.add_mqtt()
@@ -627,7 +627,8 @@ def main():
         debug_level = os.environ["DEBUG_LEVEL"]
         print(
             "Deprecated environment variable DEBUG_LEVEL "
-            "(it will be removed from version 2.x onwards)"
+            "(it will be dropped from version 2.x onwards)"
+            " please use WM_DEBUG_LEVEL instead."
         )
     except KeyError:
         debug_level = "info"
--- a/python_transport/wirepas_gateway/transport_service.py
+++ b/python_transport/wirepas_gateway/transport_service.py
@@ -23,6 +23,7 @@ from wirepas_messaging.gateway.api import (
 # This constant is the actual API level implemented by this transport module (cf WP-RM-128)
 IMPLEMENTED_API_VERSION = 1
 
+
 class TransportService(BusClient):
     """
     Implementation of gateway to backend protocol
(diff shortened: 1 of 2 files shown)


python_transport/wirepas_gateway/transport_service.py
```diff
@@ -56,15 +52,16 @@ def __init__(
...
+        ).payload
 
         self.mqtt_wrapper = MQTTWrapper(
             settings,
             self.logger,
             self._on_mqtt_wrapper_termination_cb,
             self._on_connect,
             last_will_topic,
-            last_will_message
+            last_will_message,
```
GwendalRaoul: Is it really needed?

python_transport/wirepas_gateway/transport_service.py
```diff
@@ -609,6 +627,12 @@ def main():
 
     try:
         debug_level = os.environ["DEBUG_LEVEL"]
+        print("Deprecated environment variable it will be removed with >2.x")
```
GwendalRaoul: Maybe you could name the environment variable in the message and also give the new variable name

python_transport/wirepas_gateway/utils/log_tools.py
```diff
@@ -16,7 +16,7 @@
 def setup_log(
     module,
     level="debug",
-    log_format="%(asctime)s | [%(levelname)s] %(name)s: %(message)s",
+    log_format="%(asctime)s | [%(levelname)s] %(name)s.%(funcName)s@%(filename)s:%(lineno)d: %(message)s",
```
GwendalRaoul: Log prefix will be quite long. I would remove funcName at least as filename/lineno already gives it.
But I'm not convince about filename. If name is well chosen, it should be quite straight forward to find the message in the code. No?

python_transport/wirepas_gateway/transport_service.py
PFigs: Good point

python_transport/wirepas_gateway/utils/log_tools.py
PFigs: I think the function name might be more useful as people will end up CTRL+F on all source. 

For that reason I agree that we do not need to keep the filename.

python_transport/wirepas_gateway/transport_service.py
PFigs: No, but I will let black do its thing

PFigs: I'll let black do its thing

python_transport/wirepas_gateway/utils/log_tools.py
PFigs: I removed the function name but I think the filename and number is at least needed in case the function name is reused in multiple places.

python_transport/wirepas_gateway/transport_service.py
PFigs: done

PR #114, opened 2019-08-02
Custom TLS file within transport container

### LINKED ISSUE #113: Explain how to mount a custom certificate inside a docker based installation
Add a comment in the compose files to highlight how to mount a custom certificate file inside a transport service container.

PFigs:
Clarify how to use a custom TLS file within
the transport service container.

Closes #113


README.md
```diff
@@ -335,6 +335,38 @@ or specify which container you want to view the logs from with
...
+
+```shell
+    FROM wirepas/gateway
+    COPY my-tls-file /etc/my-tls-file
+```
+
+Build the image and update the composition file to match the build name that
+you pick or use.
+
+In case you opt to mount the file within the transport service container,
```
GwendalRaoul: I would just present this second option. More option means more questions ;-) 
This second option seems enough, no?

PFigs: Agree, it is enough :+1:

PR #117, opened 2019-08-30
Avoid MQTT reconnect() loop

ansinnovation:
This endless loop is happening when the MQTT broker closes the connection early,
due to invalid credentials.
This patch aborts the MQTT processing if a disconnect is detected
before a successful connection.


PFigs:
@GwendalRaoul any comment?

PR #129, opened 2019-09-23
Add optional parameter for max poll fail duration

JanneSiren:
Add optional parameter for max poll fail duration

When debugging sink with breakpoints the sinkService dies too fast. With optional -d parameter the sinkService can be defined to stand longer poll fails. With value zero the polling will be done forever regardless of failures, though note that every poll failure is logged, thus storage capacity for log file is limiting factor.


JanneSiren:
Depends on https://github.com/wirepas/c-mesh-api/pull/17, cannot be merged to master before c-mesh-api part is merged to master.

JanneSiren:
Re run Travis CI check manually: https://travis-ci.com/wirepas/gateway/builds/128754649, but it doesn't update the check status in github. Thus using admin priviledges to merge to master.

PR #130, opened 2019-10-10
wirepas_messaging v1.3.0rc1 

JanneSiren:
This PR tests the wirepas_mesaging v1.3.0rc1


PFigs:
@GwendalRaoul I would wait for renovate to raise the update to v1.3.0 once we release it from https://github.com/wirepas/gateway. What do you think?

GwendalRaoul:
I don't really see the reason for it. 1.3.0 release of messaging doesn't include change for gateway to backend api compare to 1.2.0.
Is there a reason I missed?

PFigs:
The protobuf package has been upgraded in 1.3.0, but I dont see that as a must have upgrade. Fine for me to leave it pinned to 1.2.0

PR #132, opened 2019-10-11
Update main Readme

GwendalRaoul:
The purpose of this change is to try to clarify the different approaches to build a gateway


README.md
```diff
@@ -19,7 +19,54 @@ apis involved at each step.
 
 **Figure 1 -** Gateway services overview.
 
-## Cloning this repository
+## Instaling a Gateway
+
+Multiple options are available depending on your need.
+They are all described in the following sections.
+
+- Option 1: native installation
```
PFigs: We should add links here, similar to the TOCs we have in the other repos

README.md
```diff
@@ -19,7 +29,60 @@ apis involved at each step.
...
-## Cloning this repository
+## Installing a Gateway
+
+Multiple options are available depending on your need.
+They are all described in the following sections:
+
+[Option 1.1](#option-11)
+[Option 1.2](#option-12)
+[Option 1.1](#option-11-1)
+[Option 1.2](#option-12-1)
```
PFigs: I don't think we need these links

README.md
```diff
@@ -74,86 +137,72 @@ Dockerfiles (see [Dockerfile][here_container_dockerfile]).
...
+```xml
+    <!-- Only wirepas user can own the service name -->
+    <policy user="wirepas">
 ```
 
-## Installation
+*It is recommended to restart your gateway once this file is copied.*
 
-### Sink service
+##### Option 1.1
```
PFigs: What do you think if we have a more explicit header? Such as

Option 1.1:  from source code

README.md
GwendalRaoul: You added them right ;-) ?

README.md
GwendalRaoul: Yes sure!

README.md
PFigs: :thinking: hahah sorry for that!

README.md
```diff
@@ -19,7 +30,55 @@ apis involved at each step.
...
-## Cloning this repository
+## Installing a Gateway
+
+Multiple options are available depending on your need.
+They are all described in the following sections:
+
+- [Option 1: native installation](#option-1-native-installation)
+  - [Option 1.1](#option-11): from source code\
+    This option should be used if you plan to do modification on this reference code
+  - [Option 1.2](#option-12) from pre-built binaries\
```
PFigs: Missing ":"

README.md
```diff
@@ -74,86 +133,72 @@ Dockerfiles (see [Dockerfile][here_container_dockerfile]).
...
-In order to build the sink service, systemd headers are needed
-
-```shell
-    sudo apt install libsystemd-dev
-```
+The implementation uses system bus that has enforced security.
+In order to obtain a service name on system bus, the user launching the sink
+service must be previously declared to system.
+Provided file com.wirepas.sink.conf inside sink_service folder
+must be copied under /etc/dbus-1/system.d/ and edited with the user that will
```
PFigs: I think it would be helpful to stress the filename (italic or bold) and  
> edited with the user that will launch the sink_service

README.md
```diff
@@ -74,86 +133,72 @@ Dockerfiles (see [Dockerfile][here_container_dockerfile]).
...
+launch sink_service (and transport service).
```
PFigs: It should read, _launch the sink service_

PR #133, opened 2019-10-15
Environment arguments

### LINKED ISSUE #103: Environment variable handling within argparse
Currently we are only using one environment variable to affect the logging level, however we could and should handle it within the LoggerHelper.

My suggestion is that we handle the lookup from the host's environment in the following way:

```
       parser.add_argument(
            "--debug_level",
            action="store",
            default=os.environ.get("WM_DEBUG_LEVEL", None),
            type=str,
            help="Logger debug level",
        )
```

PFigs:
This change allows the arguments to be sourced by the
transport service directly from the environment. 

The environment parameters follow the ones used in 
backend client and wm-config v2.0.0.

Closes #103

To decide/tweak:

- [x] set final env names

### DIFF
--- a/python_transport/wirepas_gateway/utils/argument_tools.py
+++ b/python_transport/wirepas_gateway/utils/argument_tools.py
@@ -159,8 +159,8 @@ class ParserHelper:
         self.file_settings.add_argument(
             "--settings",
             type=str,
-            default=os.environ.get("WM_LXGW_FILE_SETTINGS", None),
-            default=None,
+            required=False,
+            default=os.environ.get("WM_BCLI_FILE_SETTINGS", None),
             help="A yaml file with argument parameters (see help for options).",
         )
 
--- a/python_transport/wirepas_gateway/utils/argument_tools.py
+++ b/python_transport/wirepas_gateway/utils/argument_tools.py
@@ -169,7 +169,7 @@ class ParserHelper:
         """ Commonly used MQTT arguments """
         self.mqtt.add_argument(
             "--mqtt_hostname",
-            default=os.environ.get("WM_GW_MQTT_HOST", None),
+            default=os.environ.get("WM_GW_MQTT_HOSTNAME", None),
             action="store",
             type=str,
             help="MQTT broker hostname.",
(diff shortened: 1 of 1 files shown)


python_transport/wirepas_gateway/utils/argument_tools.py
```diff
@@ -160,126 +161,132 @@ def add_file_settings(self):
...
         )
 
         self.mqtt.add_argument(
             "--mqtt_tls_version",
-            default=ssl.PROTOCOL_TLSv1_2,
+            default=os.environ.get("WM_SERVICES_MQTT_CERT_REQS", ssl.PROTOCOL_TLSv1_2),
             action="store",
             type=str,
-            help=("Specifies the version of the SSL / TLS protocol to be used."),
+            help=("Specifies the version of the  SSL / TLS protocol to be used."),
```
GwendalRaoul: really needed ;-) ?

PFigs: :man_facepalming:

python_transport/wirepas_gateway/utils/argument_tools.py
```diff
@@ -246,15 +246,15 @@ def add_mqtt(self):
 
         self.mqtt.add_argument(
             "--mqtt_tls_version",
-            default=os.environ.get("WM_SERVICES_MQTT_CERT_REQS", ssl.PROTOCOL_TLSv1_2),
+            default=os.environ.get("WM_GW_MQTT_CERT_REQS", ssl.PROTOCOL_TLSv1_2),
```
GwendalRaoul: Should it be WM_GW_TLS_VERSION?

python_transport/wirepas_gateway/utils/argument_tools.py
```diff
@@ -246,15 +246,15 @@ def add_mqtt(self):
...
             action="store",
             type=str,
             help=("Specifies the version of the SSL / TLS protocol to be used."),
         )
 
         self.mqtt.add_argument(
             "--mqtt_ciphers",
-            default=os.environ.get("WM_SERVICES_MQTT_CLIENT_KEY", None),
+            default=os.environ.get("WM_GW_MQTT_CLIENT_KEY", None),
```
GwendalRaoul: Same here

PR #139, opened 2019-10-28
Black hole

### LINKED ISSUE #29: Increase SinkCost when backhaul connection is down
When the transport service has no IP connectivity ensure the following:

* Set a maximum queue size for the MQTT queue

* Increase the sink cost based on the queue size

* Monitor the publish time of the last message

It is a good idea to allow the configuration of the queue size (if possible on the Paho https://pypi.org/project/paho-mqtt/#publishing) and publishing time.

GwendalRaoul:
Closes #29

This pull request covers the black hole issue.

WARNING: It is still under testing and may still evolve once more testing are done in a wider network

### DIFF
--- a/python_transport/wirepas_gateway/dbus/sink_manager.py
+++ b/python_transport/wirepas_gateway/dbus/sink_manager.py
@@ -94,7 +94,7 @@ class Sink:
     def _get_param(self, dic, key, attribute):
         try:
             dic[key] = getattr(self.proxy, attribute)
-        except GLib.Error:
+        except (GLib.Error, AttributeError):
             # Exception raised when getting attribute (probably not set)
             # Discard channel_map as parameter present only for old stacks
             if key != "channel_map":
--- a/python_transport/wirepas_gateway/dbus/sink_manager.py
+++ b/python_transport/wirepas_gateway/dbus/sink_manager.py
@@ -259,6 +259,25 @@ class Sink:
 
         return res
 
+    @property
+    def cost(self):
+        try:
+            cost = self.proxy.SinkCost
+        except GLib.Error:
+            self.logger.error("Cannot get sink cost for sink {}".format(self.sink_id))
+            cost = 0
+        return cost
+
+    @cost.setter
+    def cost(self, new_cost):
+        if new_cost < 0 or new_cost > 254:
+            raise ValueError("Wrong sink cost value {}".format(new_cost))
+
+        try:
+            self.proxy.SinkCost = new_cost
+        except GLib.Error:
+            self.logger.error("Cannot set sink cost for sink {}".format(self.sink_id))
+
     def get_scratchpad_status(self):
         d = {}
 
--- a/python_transport/wirepas_gateway/protocol/mqtt_wrapper.py
+++ b/python_transport/wirepas_gateway/protocol/mqtt_wrapper.py
@@ -5,8 +5,9 @@
 import queue
 import socket
 from select import select
-from threading import Thread, current_thread
+from threading import Thread
 from time import sleep
+from datetime import datetime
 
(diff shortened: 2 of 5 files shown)


python_transport/wirepas_gateway/dbus/sink_manager.py
```diff
@@ -259,6 +259,25 @@ def write_config(self, config):
...
+        try:
+            cost = self.proxy.SinkCost
+        except GLib.Error:
+            self.logger.error("Cannot get sink cost for sink {}".format(self.sink_id))
+            cost = 0
+        return cost
+
+    @cost.setter
+    def cost(self, new_cost):
+        if new_cost < 0 or new_cost > 254:
```
PFigs: Any chance it is new_cost is None? If so, the comparison will throw an exception

python_transport/wirepas_gateway/protocol/mqtt_wrapper.py
```diff
@@ -139,22 +154,24 @@ def _get_socket(self):
 
         self.logger.error("MQTT, unexpected disconnection")
 
-        # Socket is not opened anymore, try to reconnect
-        timeout = MQTTWrapper.TIMEOUT_RECONNECT_S
-        while timeout > 0:
+        # Socket is not opened anymore, try to reconnect for timeout if set
+        timeout = self.timeout
+        no_timeout = (timeout == 0)
```
PFigs: Do you think it is necessary? Python handles 0 as false.

python_transport/wirepas_gateway/protocol/mqtt_wrapper.py
```diff
@@ -139,22 +154,24 @@ def _get_socket(self):
...
+
+        while no_timeout or timeout > 0:
```
PFigs: Isn't this the same as timeout >=0?

python_transport/wirepas_gateway/protocol/mqtt_wrapper.py
```diff
@@ -200,6 +217,14 @@ def run(self):
             # thread has exited
             self.on_termination_cb()
 
+    def _publish_from_main_thread(self, topic, payload, qos, retain):
+        mid = self._client.publish(topic, payload, qos=qos, retain=retain).mid
+        # self.logger.info("Publish {} qos={}".format(mid, qos))
+        if self.publish_gueue_size == 0:
```
PFigs: Typo in property name, should read *publish_queue_size*

python_transport/wirepas_gateway/transport_service.py
```diff
@@ -621,6 +772,7 @@ def main():
     parse.add_mqtt()
     parse.add_gateway_config()
     parse.add_filtering_config()
+    parse.add_black_hole_parameters()
```
PFigs: Have you considered another name, other than black hole? I think people might not really get what the parameters are about, especially for the new ones coming into the framework. 

For example, what do you think about packet_buffer_parameters?

python_transport/wirepas_gateway/transport_service.py
```diff
@@ -73,6 +197,24 @@ def __init__(self, settings, logger=None, **kwargs):
 
         self.logger.info("Gateway started with id: %s", self.gw_id)
 
+        self.monitoringThread = None
+        self.minimun_sink_cost = settings.bh_minimal_sink_cost
+
+        if settings.bh_max_buffered_packets > 0:
+            self.logger.info(" Black hole detection enabled: max_packets=%d packets, max_delay=%d",
```
PFigs: Have you considered another name instead of black hole detection? Maybe backend connection loss would be easier to get?

PR #137, opened 2019-10-23
Update docker entrypoint

PFigs:
Docker entrypoint updates for transport and sink service environment keys.

- [x] merge transport service env handling
- [x] merge sink service env handling
- [x] review entrypoint

PR #140, opened 2019-10-28
Black hole

### LINKED ISSUE #29: Increase SinkCost when backhaul connection is down

GwendalRaoul:
Closes #29

This pull request covers the black hole issue.

In order to test some extra option must be given when starting transport module:
(extract of wm-gw --help)
 buffering:
  --buffering_max_buffered_packets BUFFERING_MAX_BUFFERED_PACKETS
                        Maximum number of messages to buffer before rising
                        sink cost (0 will disable feature) (default: 0)
  --buffering_max_delay_without_publish BUFFERING_MAX_DELAY_WITHOUT_PUBLISH
                        Maximum time to wait in seconds without any successful
                        publish with packet queued before rising sink cost (0
                        will disable feature) (default: 0)
  --buffering_minimal_sink_cost BUFFERING_MINIMAL_SINK_COST
                        Minimal sink cost for a sink on this gateway. Can be
                        used to minimize traffic on a gateway, but it will
                        reduce maximum number of hops for this gateway
                        (default: 0)

### DIFF
--- a/python_transport/wirepas_gateway/dbus/sink_manager.py
+++ b/python_transport/wirepas_gateway/dbus/sink_manager.py
@@ -280,6 +280,25 @@ class Sink:
 
         return res
 
+    @property
+    def cost(self):
+        try:
+            cost = self.proxy.SinkCost
+        except GLib.Error:
+            self.logger.error("Cannot get sink cost for sink {}".format(self.sink_id))
+            cost = 0
+        return cost
+
+    @cost.setter
+    def cost(self, new_cost):
+        if new_cost is None or new_cost < 0 or new_cost > 254:
+            raise ValueError("Wrong sink cost value {}".format(new_cost))
+
+        try:
+            self.proxy.SinkCost = new_cost
+        except GLib.Error:
+            self.logger.error("Cannot set sink cost for sink {}".format(self.sink_id))
+
     def get_scratchpad_status(self):
         d = {}
 
--- a/python_transport/wirepas_gateway/protocol/mqtt_wrapper.py
+++ b/python_transport/wirepas_gateway/protocol/mqtt_wrapper.py
@@ -23,9 +23,6 @@ class MQTTWrapper(Thread):
     # Keep alive time with broker
     KEEP_ALIVE_S = 60
 
-    # Reconnect timeout in Seconds
-    TIMEOUT_RECONNECT_S = 120
-
     def __init__(
         self,
         settings,
@@ -75,6 +72,8 @@ class MQTTWrapper(Thread):
         except (socket.gaierror, ValueError) as e:
             self.logger.error("Cannot connect to mqtt %s", e)
             exit(-1)
+
+        self.timeout = settings.mqtt_reconnect_delay
 
         # Set options to initial socket
         self._client.socket().setsockopt(socket.SOL_SOCKET, socket.SO_SNDBUF, 2048)
@@ -149,24 +148,29 @@ class MQTTWrapper(Thread):
             self.logger.error("Impossible to connect - authentication failure ?")
             return None
 
-        # Socket is not opened anymore, try to reconnect
-        timeout = MQTTWrapper.TIMEOUT_RECONNECT_S
-        while timeout > 0:
+        # Socket is not opened anymore, try to reconnect for timeout if set
+        loop_forever = (self.timeout == 0)
+        delay = 0
+
+        # Loop forever or until timeout is over
+        while loop_forever or (delay <= self.timeout):
             try:
                 ret = self._client.reconnect()
                 if ret == mqtt.MQTT_ERR_SUCCESS:
                     break
             except Exception:
-                # Retry to connect in 1 sec up to timeout
+                # Retry to connect in 1 sec up to timeout if set
                 sleep(1)
-                timeout -= 1
+                delay += 1
                 self.logger.debug("Retrying to connect in 1 sec")
 
-        if timeout <= 0:
-            self.logger.error(
-                "Unable to reconnect after %s seconds", MQTTWrapper.TIMEOUT_RECONNECT_S
-            )
-            return None
+        if not loop_forever:
+            # In case of timeout set, check if it exits because of timeout
+            if delay > self.timeout:
+                self.logger.error(
+                    "Unable to reconnect after %s seconds", delay
+                )
+                return None
 
         # Socket must be available once reconnect is successful
         if self._client.socket() is None:
--- a/python_transport/wirepas_gateway/protocol/mqtt_wrapper.py
+++ b/python_transport/wirepas_gateway/protocol/mqtt_wrapper.py
@@ -228,6 +228,14 @@ class MQTTWrapper(Thread):
             self.on_termination_cb()
 
     def _publish_from_wrapper_thread(self, topic, payload, qos, retain):
+        """
+        Internal method to publish on Mqtt. This method is only called from
+        mqtt wrapper thread to avoid races.
+        :param topic: Topic to publish on
+        :param payload: Payload
+        :param qos: Qos to use
+        :param retain: Is it a retain message
+        """
         mid = self._client.publish(topic, payload, qos=qos, retain=retain).mid
         # self.logger.info("Publish {} qos={}".format(mid, qos))
         if self.publish_queue_size == 0:
@@ -237,7 +245,7 @@ class MQTTWrapper(Thread):
 
     def publish(self, topic, payload, qos=1, retain=False) -> None:
         """
-        Method to publish to Mqtt from a different thread.
+        Method to publish to Mqtt from any thread.
         :param topic: Topic to publish on
         :param payload: Payload
         :param qos: Qos to use
--- a/python_transport/wirepas_gateway/protocol/mqtt_wrapper.py
+++ b/python_transport/wirepas_gateway/protocol/mqtt_wrapper.py
@@ -6,8 +6,9 @@ import queue
 import socket
 import ssl
 from select import select
-from threading import Thread, current_thread
+from threading import Thread
 from time import sleep
+from datetime import datetime
 
 from paho.mqtt import client as mqtt
 from paho.mqtt.client import connack_string
@@ -38,11 +39,16 @@ class MQTTWrapper(Thread):
         self.logger = logger
         self.on_termination_cb = on_termination_cb
         self.on_connect_cb = on_connect_cb
+        # Set to track the unpublished packets
+        self._unpublished_mid_set = set()
+        # Variable to keep track of latest published packet
+        self._timestamp_last_publish = datetime.now()
 
         self._client = mqtt.Client(
             client_id=settings.gateway_id,
             clean_session=not settings.mqtt_persist_session,
         )
+
         if not settings.mqtt_force_unsecure:
             try:
                 self._client.tls_set(
@@ -59,6 +65,7 @@ class MQTTWrapper(Thread):
 
         self._client.username_pw_set(settings.mqtt_username, settings.mqtt_password)
         self._client.on_connect = self._on_connect
+        self._client.on_publish = self._on_publish
 
         if last_will_topic is not None and last_will_data is not None:
             self._set_last_will(last_will_topic, last_will_data)
@@ -95,6 +102,11 @@ class MQTTWrapper(Thread):
         if self.on_connect_cb is not None:
             self.on_connect_cb()
 
+    def _on_publish(self, client, userdata, mid):
+        self._unpublished_mid_set.remove(mid)
+        self._timestamp_last_publish = datetime.now()
+        return
+
     def _do_select(self, sock):
         # Select with a timeout of 1 sec to call loop misc from time to time
         r, w, _ = select(
@@ -111,7 +123,10 @@ class MQTTWrapper(Thread):
                 # next select will exit immediately if queue not empty
                 while True:
                     topic, payload, qos, retain = self._publish_queue.get()
-                    self._client.publish(topic, payload, qos=qos, retain=retain)
+
+                    self._publish_from_wrapper_thread(
+                        topic, payload, qos=qos, retain=retain
+                    )
 
                     # FIX: read internal sockpairR as it is written but
                     # never read as we don't use the internal paho loop
@@ -149,7 +164,7 @@ class MQTTWrapper(Thread):
             return None
 
         # Socket is not opened anymore, try to reconnect for timeout if set
-        loop_forever = (self.timeout == 0)
+        loop_forever = self.timeout == 0
         delay = 0
 
         # Loop forever or until timeout is over
@@ -167,9 +182,7 @@ class MQTTWrapper(Thread):
         if not loop_forever:
             # In case of timeout set, check if it exits because of timeout
             if delay > self.timeout:
-                self.logger.error(
-                    "Unable to reconnect after %s seconds", delay
-                )
+                self.logger.error("Unable to reconnect after %s seconds", delay)
                 return None
 
         # Socket must be available once reconnect is successful
@@ -214,6 +227,14 @@ class MQTTWrapper(Thread):
             # thread has exited
             self.on_termination_cb()
 
+    def _publish_from_wrapper_thread(self, topic, payload, qos, retain):
+        mid = self._client.publish(topic, payload, qos=qos, retain=retain).mid
+        # self.logger.info("Publish {} qos={}".format(mid, qos))
+        if self.publish_queue_size == 0:
+            # Reset last published packet
+            self._timestamp_last_publish = datetime.now()
+        self._unpublished_mid_set.add(mid)
+
     def publish(self, topic, payload, qos=1, retain=False) -> None:
         """
         Method to publish to Mqtt from a different thread.
@@ -222,17 +243,22 @@ class MQTTWrapper(Thread):
         :param qos: Qos to use
         :param retain: Is it a retain message
         """
-        if current_thread().ident == self.ident:
-            # Already on right thread
-            self._client.publish(topic, payload, qos, retain)
-        else:
-            # Send it to the queue to be published from Mqtt thread
-            self._publish_queue.put((topic, payload, qos, retain))
+        # Send it to the queue to be published from Mqtt thread
+        self._publish_queue.put((topic, payload, qos, retain))
 
     def subscribe(self, topic, cb, qos=2) -> None:
         self.logger.debug("Subscribing to: {}".format(topic))
         self._client.subscribe(topic, qos)
         self._client.message_callback_add(topic, cb)
+
+    @property
+    def publish_queue_size(self):
+        return len(self._unpublished_mid_set)
+
+    @property
+    def last_published_packet_s(self):
+        delta = datetime.now() - self._timestamp_last_publish
+        return delta.total_seconds()
 
 
 class SelectableQueue(queue.Queue):
--- a/python_transport/wirepas_gateway/protocol/mqtt_wrapper.py
+++ b/python_transport/wirepas_gateway/protocol/mqtt_wrapper.py
@@ -228,28 +228,31 @@ class MQTTWrapper(Thread):
             self.on_termination_cb()
 
     def _publish_from_wrapper_thread(self, topic, payload, qos, retain):
-        """
-        Internal method to publish on Mqtt. This method is only called from
+        """Internal method to publish on Mqtt. This method is only called from
         mqtt wrapper thread to avoid races.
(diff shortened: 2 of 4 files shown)


python_transport/wirepas_gateway/protocol/mqtt_wrapper.py
```diff
@@ -40,8 +38,13 @@ def __init__(
         self.logger = logger
         self.on_termination_cb = on_termination_cb
         self.on_connect_cb = on_connect_cb
+        # Set to track the unpublished packets
+        self._unpublished_mid_set = set()
+        # Variable to keep track of latest published packet
+        self._timestamp_last_publish = datetime.now()
```
PFigs: Should the timestamp be time zone aware? You can set one, eg, utc by setting the tz argument https://docs.python.org/3.7/library/datetime.html#datetime.datetime.now

I've run into issues in the past where a UTC timestamp was interpret as local time, since the +00 was left out. But that was due to the datetime -> isoformat -> datetime. 

If this is left as unix timestamp, then it is okay as is.

GwendalRaoul: I don't think it makes any difference. It is mainly internal use and not exported. As long as all timestamp are taken the same way, it shouldn't matter as we just care about delta

PFigs: Agree. Safe to leave it as is.

python_transport/wirepas_gateway/dbus/sink_manager.py
```diff
@@ -259,6 +259,25 @@ def write_config(self, config):
...
+        try:
+            cost = self.proxy.SinkCost
+        except GLib.Error:
+            self.logger.error("Cannot get sink cost for sink {}".format(self.sink_id))
+            cost = 0
+        return cost
+
+    @cost.setter
+    def cost(self, new_cost):
+        if new_cost < 0 or new_cost > 254:
```
PFigs: If new cost is None, the comparison will throw an error

python_transport/wirepas_gateway/protocol/mqtt_wrapper.py
```diff
@@ -145,22 +160,24 @@ def _get_socket(self):
...
             return None
 
-        # Socket is not opened anymore, try to reconnect
-        timeout = MQTTWrapper.TIMEOUT_RECONNECT_S
-        while timeout > 0:
+        # Socket is not opened anymore, try to reconnect for timeout if set
+        timeout = self.timeout
+        no_timeout = (timeout == 0)
+
+        while no_timeout or timeout > 0:
```
PFigs: Am I wrong or could this comparison be simplified further?

`while timeout ==0 or timeout >0` => `while timeout =>0`

python_transport/wirepas_gateway/protocol/mqtt_wrapper.py
```diff
@@ -145,22 +160,24 @@ def _get_socket(self):
...
                     break
             except Exception:
-                # Retry to connect in 1 sec up to timeout
+                # Retry to connect in 1 sec up to timeout if set
                 sleep(1)
                 timeout -= 1
                 self.logger.debug("Retrying to connect in 1 sec")
 
-        if timeout <= 0:
+        if not no_timeout and timeout <= 0:
```
PFigs: Similar thing here, isnt this the same as 

`if timeout < 0 :`

python_transport/wirepas_gateway/protocol/mqtt_wrapper.py
```diff
@@ -206,6 +223,14 @@ def run(self):
             # thread has exited
             self.on_termination_cb()
 
+    def _publish_from_main_thread(self, topic, payload, qos, retain):
```
PFigs: Missing doc string

python_transport/wirepas_gateway/protocol/mqtt_wrapper.py
```diff
@@ -206,6 +223,14 @@ def run(self):
...
+        mid = self._client.publish(topic, payload, qos=qos, retain=retain).mid
+        # self.logger.info("Publish {} qos={}".format(mid, qos))
+        if self.publish_gueue_size == 0:
```
PFigs: Typo in attribute name, it should read self.publish_queue_size

python_transport/wirepas_gateway/protocol/mqtt_wrapper.py
```diff
@@ -214,18 +239,22 @@ def publish(self, topic, payload, qos=1, retain=False) -> None:
...
+        # Send it to the queue to be published from Mqtt thread
+        self._publish_queue.put((topic, payload, qos, retain))
 
     def subscribe(self, topic, cb, qos=2) -> None:
         self.logger.debug("Subscribing to: {}".format(topic))
         self._client.subscribe(topic, qos)
         self._client.message_callback_add(topic, cb)
 
+    @property
+    def publish_gueue_size(self):
```
PFigs: Same typo here (queue)

python_transport/wirepas_gateway/transport_service.py
```diff
@@ -27,6 +27,127 @@
...
+class ConnectionToBackendMonitorThread(Thread):
+
+    # Maximum cost to disable traffic
+    SINK_COST_HIGH = 254
+
+    """
+    Thread monitoring the connection with the broker.
+    Connection status is monitored thanks to the number of
+    messages pushed to the mqtt client queue but not published
+    and the latest successfully published packet
```
PFigs: Missing dot.

python_transport/wirepas_gateway/transport_service.py
```diff
@@ -27,6 +27,127 @@
 IMPLEMENTED_API_VERSION = 1
 
 
+class ConnectionToBackendMonitorThread(Thread):
+
+    # Maximum cost to disable traffic
+    SINK_COST_HIGH = 254
+
+    """
+    Thread monitoring the connection with the broker.
```
PFigs: MQTT broker

python_transport/wirepas_gateway/transport_service.py
```diff
@@ -27,6 +27,127 @@
...
+    Connection status is monitored thanks to the number of
+    messages pushed to the mqtt client queue but not published
+    and the latest successfully published packet
+    Mqtt connection is not used as a trigger as it may modify too often
+    the sink cost with an unstable connection that would be counterproductive
```
PFigs: Missing dot here as well

python_transport/wirepas_gateway/transport_service.py
```diff
@@ -27,6 +27,127 @@
...
+
+    It could also be implemented without a periodic check and tested
```
PFigs: Is this paragraph needed?

python_transport/wirepas_gateway/transport_service.py
```diff
@@ -27,6 +27,127 @@
...
+    each time a packet is published but on performance point of view it would
+    be the same
+    """
+    def __init__(self, logger, period, mqtt_wrapper, sink_manager,
+                 minimum_sink_cost, max_buffered_packets, max_delay_without_publish):
```
PFigs: Could you explain the attributes in the Class's doc string? 

I think the max_buffered_packets should have a minimum value.

python_transport/wirepas_gateway/transport_service.py
```diff
@@ -27,6 +27,127 @@
...
+                    self.logger.info("Increasing sink cost of all sinks")
+                    self.logger.debug("Last publish: %s Queue Size %s",
+                                      self.mqtt_wrapper.last_published_packet_s,
+                                      self.mqtt_wrapper.publish_gueue_size)
+
+                    self._set_sinks_cost_high()
+                    self.disconnected = True
+            else:
+                if self.mqtt_wrapper.publish_gueue_size == 0:
+                    # Network is back, put the connection back
```
PFigs: I think it could take a while for the buffer queue to reach zero, especially if it has been queuing up packets for a while. Aslo, could we end up in a race condition where the queue exhaust is not sufficient to compensate the consumption of network packets?

python_transport/wirepas_gateway/transport_service.py
```diff
@@ -73,6 +197,24 @@ def __init__(self, settings, logger=None, **kwargs):
 
         self.logger.info("Gateway started with id: %s", self.gw_id)
 
+        self.monitoringThread = None
+        self.minimun_sink_cost = settings.bh_minimal_sink_cost
+
+        if settings.bh_max_buffered_packets > 0:
+            self.logger.info(" Black hole detection enabled: max_packets=%d packets, max_delay=%d",
```
PFigs: Have you considered a different name for black hole detection? I think it would be clearer if we picked, something like connection loss or connection broken.

python_transport/wirepas_gateway/transport_service.py
```diff
@@ -621,6 +772,7 @@ def main():
     parse.add_mqtt()
     parse.add_gateway_config()
     parse.add_filtering_config()
+    parse.add_black_hole_parameters()
```
PFigs: Same comment as above regarding the choice of name

python_transport/wirepas_gateway/protocol/mqtt_wrapper.py
GwendalRaoul: No, it is maybe not well written but timeout and no_timeout is really different.

In fact timeout is initialized to self.timeout (being the max value set by user) and is updated in the loop.
But in case self.timeout is 0 (meaning reconnect forever), timeout is bypassed and it is a while True.

But as you raised the point, I will try to make it more clear to avoid confusion

python_transport/wirepas_gateway/dbus/sink_manager.py
GwendalRaoul: Done

python_transport/wirepas_gateway/protocol/mqtt_wrapper.py
GwendalRaoul: Same reply, no :)

I definitely have to write it differently.

python_transport/wirepas_gateway/protocol/mqtt_wrapper.py
GwendalRaoul: Done

python_transport/wirepas_gateway/transport_service.py
GwendalRaoul: You mean a maximum value? an upper bound? So if not defined by user it is 0 and right now the algorithm is disabled and messages are buffered forever.
What would be the maximum value?
I think the correct solution would be to define a maximum and start overriding old messages if limit is reached but will not be implemented in this change.
I guess for now, this upper limit can be defined by user.

PFigs: I meant maximum value 😅 got a bit mixed up with the parameter name. I think a few k messages would be good and as you said, starting to pop old messages to make room for new ones.

Fine with me if we tackle this in a future optimization.

python_transport/wirepas_gateway/protocol/mqtt_wrapper.py
PFigs: Thanks :cookie:

python_transport/wirepas_gateway/transport_service.py
GwendalRaoul: I'm not sure to get your point.
Yes it may take a while (depending on the maximum number of packets set), and that's why I prefer to flush everything before reducing the sink cost again.
I don't really understand the race you mentioned. The thread check periodically the status of queue so at some point the queue will be empty.

python_transport/wirepas_gateway/transport_service.py
GwendalRaoul: so add_backend_connection_loss_parameters()

and parameters are prefixed with bcl_ ?

python_transport/wirepas_gateway/transport_service.py
GwendalRaoul: Not really.
What would be your preference:
BackendConnectionLoss?

PFigs: Would be a sensible choice in my opinion

python_transport/wirepas_gateway/transport_service.py
PFigs: thumbs up for add_backend_connection_loss_parameters(). You could even simplify it further, such as connection_loss_settings() or connection_buffer_settings().


For the arguments, I would prefer a longer name. I think it is always easier to remember, such as *connection_* or *buffer_*

python_transport/wirepas_gateway/transport_service.py
PFigs: I was aiming at a situation where we try to fill and exhaust the queue at the same time, thus we could end up in a situation where we could not reach zero. But if you guarantee that the queue is exhausted first, then it is okay :+1:

PFigs:
Codacy Here is an overview of what got changed by this pull request:


```diff

Complexity increasing per file
==============================
- python_transport/wirepas_gateway/protocol/mqtt_wrapper.py  2
         
```


See the complete overview on [Codacy](https://app.codacy.com/app/gateway/gateway/pullRequest?prid=4394465&bid=14897242)

python_transport/wirepas_gateway/protocol/mqtt_wrapper.py
```diff
@@ -210,26 +227,47 @@ def run(self):
             # thread has exited
             self.on_termination_cb()
 
+    def _publish_from_wrapper_thread(self, topic, payload, qos, retain):
+        """
+        Internal method to publish on Mqtt. This method is only called from
+        mqtt wrapper thread to avoid races.
+        :param topic: Topic to publish on
```
PFigs: Please use [this style](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html) for the doc strings since it is the one we _try_ to observe

python_transport/wirepas_gateway/protocol/mqtt_wrapper.py
```diff
@@ -210,26 +227,47 @@ def run(self):
...
+        :param payload: Payload
+        :param qos: Qos to use
+        :param retain: Is it a retain message
+        """
+        mid = self._client.publish(topic, payload, qos=qos, retain=retain).mid
+        # self.logger.info("Publish {} qos={}".format(mid, qos))
```
PFigs: please remove this comment

python_transport/wirepas_gateway/protocol/mqtt_wrapper.py
```diff
@@ -210,26 +227,47 @@ def run(self):
...
         :param retain: Is it a retain message
         """
-        if current_thread().ident == self.ident:
-            # Already on right thread
-            self._client.publish(topic, payload, qos, retain)
-        else:
-            # Send it to the queue to be published from Mqtt thread
-            self._publish_queue.put((topic, payload, qos, retain))
+        # Send it to the queue to be published from Mqtt thread
+        self._publish_queue.put((topic, payload, qos, retain))
```
PFigs: Would a dict be easier to handle? I see that we might easily make a mistake accessing the wrong index on the other side of the queue

python_transport/wirepas_gateway/transport_service.py
```diff
@@ -73,6 +217,27 @@ def __init__(self, settings, logger=None, **kwargs):
 
         self.logger.info("Gateway started with id: %s", self.gw_id)
 
+        self.monitoringThread = None
```
PFigs: Could you avoid using camelCase for this variable please?

python_transport/wirepas_gateway/utils/argument_tools.py
```diff
@@ -290,6 +290,56 @@ def add_mqtt(self):
             help=("When true the client will skip the TLS check."),
         )
 
+        self.mqtt.add_argument(
+            "--mqtt_reconnect_delay",
+            default=0,
```
PFigs: Could we add env handling in here as well? Unless we merge https://github.com/wirepas/gateway/pull/137 afterwards

python_transport/wirepas_gateway/protocol/mqtt_wrapper.py
GwendalRaoul: I don't think so. The queue is only internal and the tuple seems fine to me. Anyone using this queue must know exactly what it contains. I see a dictionary here a bit overkill ;-)

BTW, it is not related to this change ;-)

python_transport/wirepas_gateway/transport_service.py
GwendalRaoul: Sure!!
Could we enforce it to avoid this kind of mistake?

python_transport/wirepas_gateway/utils/argument_tools.py
GwendalRaoul: Yes, it depends in which order we merge them.
What about the other "buffering" settings? I guess we should add one also for them

PFigs:
Codacy Here is an overview of what got changed by this pull request:


```diff

Complexity increasing per file
==============================
- python_transport/wirepas_gateway/protocol/mqtt_wrapper.py  2
         
```


See the complete overview on [Codacy](https://app.codacy.com/app/wirepas/gateway/pullRequest?prid=4394465&bid=14897242)

python_transport/wirepas_gateway/protocol/mqtt_wrapper.py
```diff
@@ -149,24 +163,27 @@ def _get_socket(self):
...
-        while timeout > 0:
+        # Socket is not opened anymore, try to reconnect for timeout if set
+        loop_forever = self.timeout == 0
+        delay = 0
+
+        # Loop forever or until timeout is over
+        while loop_forever or (delay <= self.timeout):
             try:
                 ret = self._client.reconnect()
                 if ret == mqtt.MQTT_ERR_SUCCESS:
```
PFigs: I think we could improve this on a separate PR, for example, I think there is no point to retry if the error signals an auth issue.

python_transport/wirepas_gateway/protocol/mqtt_wrapper.py
PFigs: Fine with me :+1:

python_transport/wirepas_gateway/utils/argument_tools.py
PFigs: Yes, it applies to the others as well. 

Lets tackle it on #137, makes more sense since it is not yet merged.

python_transport/wirepas_gateway/transport_service.py
```diff
@@ -73,6 +217,27 @@ def __init__(self, settings, logger=None, **kwargs):
 
         self.logger.info("Gateway started with id: %s", self.gw_id)
 
+        self.monitoring_thread = None
+        self.minimum_sink_cost = settings.buffering_minimal_sink_cost
+
+        if settings.buffering_max_buffered_packets > 0:
+            self.logger.info(
+                " Black hole detection enabled: max_packets=%d packets, max_delay=%d",
```
PFigs: I think %s would be safer in here

python_transport/wirepas_gateway/protocol/mqtt_wrapper.py
GwendalRaoul: It should already be handled as it will not loop if there were no successful connection yet (probably the case if auth isue).
But agree that it can be improved

python_transport/wirepas_gateway/transport_service.py
GwendalRaoul: Done

PFigs:
Codacy Here is an overview of what got changed by this pull request:


```diff

Complexity increasing per file
==============================
- python_transport/wirepas_gateway/protocol/mqtt_wrapper.py  2
         
```


See the complete overview on [Codacy](https://app.codacy.com/app/wirepas/gateway/pullRequest?prid=4394465&bid=14897242)

PFigs:
Codacy Here is an overview of what got changed by this pull request:


```diff

Complexity increasing per file
==============================
- python_transport/wirepas_gateway/protocol/mqtt_wrapper.py  2
         
```


See the complete overview on [Codacy](https://app.codacy.com/app/wirepas/gateway/pullRequest?prid=4394465&bid=14897242)

PR #144, opened 2019-11-05
Fix parsing of cert_reqs and tls_version settings

### LINKED ISSUE #143: [BUG]  Some SSL settings are not parsed correctly from the commandline
Some SSL settings parsed in `argument_tools.py` for the gateway are parsed as strings, but are `enum`s backed by `int`s within the code. The specific settings affected are `mqtt_cert_reqs` and `mqtt_tls_version`. Entering any value for these settings (changing them from the default) produces an error in my experience.

A simple fix would be to change the type to `int` and pass the appropriate value for the enum. imo a better fix (that isn't reliant on knowing the int value of the enum) would be to continue to take strings, filling in the `add_argument` `choices` parameter with the possible values, and cast the setting from string when they are used (in `mqtt_wrapper.py`)

skirkham:
Closes #143 .

Allows parsing a `cert_reqs` and `tls_version` settings from the command line.


python_transport/wirepas_gateway/utils/argument_tools.py
```diff
@@ -241,7 +242,13 @@ def add_mqtt(self):
 
         self.mqtt.add_argument(
             "--mqtt_tls_version",
-            default=ssl.PROTOCOL_TLSv1_2,
+            default="PROTOCOL_TLSv1_2",
+            choices=["PROTOCOL_TLS",
```
GwendalRaoul: It looks like using PROTOCOL_TLSvx is deprecated: https://docs.python.org/3/library/ssl.html#ssl.PROTOCOL_TLSv1
So we should use PROTOCOL_TLS and then use OP_* flags.

But it is totally out of scope for this change and will be addressed in a different PR.

Thanks for your contribution!

PR #149, opened 2019-11-21
Fix overflow

### LINKED ISSUE #147: [BUG]  Check value range before Dbus call
**Describe the bug**
Most values come from protobuf as uint32 (as there is no uint8 or uint16).
But some of them are interpreted as uint8 or uint16 on dbus and giving a value ouside of the range will generate an exception

**To Reproduce**
Try to set app config with seq = 260 for example

**Expected behavior**
Get an invalid_param error

GwendalRaoul:
Closes #147

### DIFF
--- a/python_transport/wirepas_gateway/dbus/sink_manager.py
+++ b/python_transport/wirepas_gateway/dbus/sink_manager.py
@@ -259,7 +259,13 @@ class Sink:
             new_state = stack_started
 
         try:
-            self.proxy.SetStackState(new_state)
+            current_state = (self.proxy.StackStatus & 0x01) == 0
+            # Change stack state only if needed to avoid unnecessary events
+            if current_state != new_state:
+                self.logger.debug(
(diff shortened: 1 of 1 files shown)


python_transport/wirepas_gateway/dbus/sink_manager.py
```diff
@@ -244,7 +259,13 @@ def write_config(self, config):
             new_state = stack_started
 
         try:
-            self.proxy.SetStackState(new_state)
+            current_state = (self.proxy.StackStatus & 0x01) == 0
```
PFigs: What is the reason behind the and with 0x01?

GwendalRaoul: It's because StackStatus is a bitfield with bit 0 being StackRunning(0) or stackStopped(1).
So current_state is True if started and false otherwise

PFigs: Thanks!

PR #150, opened 2019-11-27
Fix: fix format from c extension to python

GwendalRaoul:
Format was signed value instead of unsigned.
It had no impact until now as there was no uint32 with highest
bit set to one, but it is now possible with 32 bit unicast address

Without change, python will consider it as a negative value,
and the it will be signed extended on more than 32 bits creating
an overflow when using it in protobuf


PFigs:
Gave it a try on my desk and I can confirm the change fixes the address issue. Merge away captain! :boat:

PR #161, opened 2020-01-16
Integer handling

PFigs:
Integer arguments are not being converted to integers when routed through the
host's environment.

### DIFF
--- a/python_transport/wirepas_gateway/utils/argument_tools.py
+++ b/python_transport/wirepas_gateway/utils/argument_tools.py
@@ -161,7 +161,7 @@ class ParserHelper:
             return value
         if value.lower() in ("yes", "true", "t", "y", "1"):
             return True
-        elif value.lower() in ("no", "false", "f", "n", "0"):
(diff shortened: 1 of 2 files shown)


python_transport/wirepas_gateway/utils/argument_tools.py
```diff
@@ -166,6 +166,16 @@ def str2bool(value):
         else:
             raise argparse.ArgumentTypeError("Boolean value expected.")
 
+    @staticmethod
+    def str2int(value):
+        """ Ensures string to bool conversion """
+        try:
+            value = int(value)
+        except ValueError:
+            value = 0
```
GwendalRaoul: I would have keep the exception.

If you do a typo for port to 1883a for example, it will default the port to 0 and silently try a connection on this port...

PFigs: Good point. I'll add it back as it is keeps things more predictable.

PR #167, opened 2020-01-17
Send data debug verbosity

PFigs:
Increases the verbosity of the send data.

It is of particular importance to have the event id, 
so one can correlate it on the consumer side.

### DIFF
--- a/python_transport/wirepas_gateway/transport_service.py
+++ b/python_transport/wirepas_gateway/transport_service.py
@@ -346,9 +346,7 @@ class TransportService(BusClient):
         topic = TopicGenerator.make_received_data_topic(
             self.gw_id, sink_id, network_address, src_ep, dst_ep
         )
-        self.logger.debug(
-            "Send data: %s | %s", topic, event.event_id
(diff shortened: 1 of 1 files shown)


python_transport/wirepas_gateway/transport_service.py
```diff
@@ -346,7 +346,7 @@ def on_data_received(
         topic = TopicGenerator.make_received_data_topic(
             self.gw_id, sink_id, network_address, src_ep, dst_ep
         )
-        self.logger.debug("Sending data to: %s", topic)
+        self.logger.debug("send data: %s | %s | %s | %s", topic, event.event_id, data, data_size)
```
GwendalRaoul: Send instead of send?
I'm not a huge fan to print the data. Data size is fine but content doesn't make a lot of sense to me.

PFigs: Caps is fine with me. I think that is how the other print are, so let's be consistent. We can drop the data, the event id and topic is enough.

PR #169, opened 2020-02-10
mqtt_wrapper: fix buffering issues during reconnect

ansinnovation:
first issue: while reconnecting, the wrapper thread never
read the internal Selectable Queue. The queue size is not
limited BUT the socket used for signaling has a memory limit,
and will start to be blocking after (approximately) 400
published messages.

The patch change the way the signal socket is used.

second issue: The messages stored in the internal Selectable Queue
are not used for the publish queue size computation. It prevents the
black hole detection to work while reconnection is in progress.

The patch update the publish queue size computation.

### DIFF
--- a/python_transport/wirepas_gateway/protocol/mqtt_wrapper.py
+++ b/python_transport/wirepas_gateway/protocol/mqtt_wrapper.py
@@ -6,7 +6,7 @@ import queue
 import socket
 import ssl
 from select import select
-from threading import Thread
+from threading import Thread, Lock
 from time import sleep
 from datetime import datetime
 
@@ -264,7 +264,7 @@ class MQTTWrapper(Thread):
 
     @property
     def publish_queue_size(self):
-        return len(self._unpublished_mid_set)
+        return len(self._unpublished_mid_set) + self._publish_queue.get_size()
 
     @property
     def last_published_packet_s(self):
@@ -281,6 +281,8 @@ class SelectableQueue(queue.Queue):
     def __init__(self):
         super().__init__()
         self._putsocket, self._getsocket = socket.socketpair()
+        self._lock = Lock()
+        self._size = 0
 
     def fileno(self):
         """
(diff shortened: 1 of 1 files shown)


python_transport/wirepas_gateway/protocol/mqtt_wrapper.py
```diff
@@ -290,15 +292,20 @@ def fileno(self):
         return self._getsocket.fileno()
 
     def put(self, item, block=True, timeout=None):
-        # Insert item in queue
-        super().put(item, block, timeout)
-        # Send 1 byte on socket to signal select
-        self._putsocket.send(b"x")
+        with self._lock:
+            if self.empty():
```
GwendalRaoul: From the queue documentation:

_Return True if the queue is empty, False otherwise (not reliable!)._
_This method is likely to be removed at some point._

Any idea why it is not reliable and what could be the consequences?

As an alternative, I would keep the size of the queue as an attribute of this child class. We would have the control of it instead of relaying on the empty() implementation.

python_transport/wirepas_gateway/protocol/mqtt_wrapper.py
```diff
@@ -7,6 +7,7 @@
 import ssl
 from select import select
 from threading import Thread
+from threading import Lock
```
GwendalRaoul: from threading import Thread, Lock

python_transport/wirepas_gateway/protocol/mqtt_wrapper.py
ansinnovation: My understanding is that the "not reliable" is a warning for multi-threaded issues.
The doc (https://docs.python.org/3.8/library/queue.html) says:
Return True if the queue is empty, False otherwise. If empty() returns True it doesn’t guarantee that a subsequent call to put() will not block. Similarly, if empty() returns False it doesn’t guarantee that a subsequent call to get() will not block.

And there is not mention of a deprecation.

It's hard to believe that a Queue implementation is not able to provide the number of elements...

python_transport/wirepas_gateway/protocol/mqtt_wrapper.py
```diff
@@ -264,7 +264,7 @@ def subscribe(self, topic, cb, qos=2) -> None:
 
     @property
     def publish_queue_size(self):
-        return len(self._unpublished_mid_set)
+        return len(self._unpublished_mid_set) + self._publish_queue.qsize()
```
GwendalRaoul: Why not using _size instead? With a get_size method from SelectableQueue

GwendalRaoul:
I tested the patch on my side too and it looks good to me.
Thanks for the good contribution!

PR #210, opened 2020-12-15
Wirepas messaging: remove need for wirepas_messaging

GwendalRaoul:
Use new wireps_mesh_messaging wheel that is targeting only
the gateway to/from backend api compare to old wirepas_messaging


GwendalRaoul:
Closed as already merged from other PR

PR #223, opened 2021-06-08
Adding error test and log messages in mqtt wrapper

ansinnovation:
This code was used to diagnose the race issue in the dbus client.
It should make sense to include it to debug issues with
MQTT connections/brokers

### DIFF
--- a/python_transport/wirepas_gateway/protocol/mqtt_wrapper.py
+++ b/python_transport/wirepas_gateway/protocol/mqtt_wrapper.py
@@ -79,6 +79,7 @@ class MQTTWrapper(Thread):
         self._client.username_pw_set(settings.mqtt_username, settings.mqtt_password)
         self._client.on_connect = self._on_connect
         self._client.on_publish = self._on_publish
+        self._client.on_disconnect = self._on_disconnect
 
         if last_will_topic is not None and last_will_data is not None:
             self._set_last_will(last_will_topic, last_will_data)
@@ -90,7 +91,13 @@ class MQTTWrapper(Thread):
                 keepalive=MQTTWrapper.KEEP_ALIVE_S,
             )
         except (socket.gaierror, ValueError) as e:
-            self.logger.error("Cannot connect to mqtt %s", e)
+            self.logger.error(
+                "Error on MQTT address %s:%d => %s"
+                % (settings.mqtt_hostname, settings.mqtt_port, str(e))
+            )
+            exit(-1)
+        except ConnectionRefusedError:
+            self.logger.error("Connection Refused by MQTT broker")
             exit(-1)
 
         self.timeout = settings.mqtt_reconnect_delay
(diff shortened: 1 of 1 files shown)


python_transport/wirepas_gateway/protocol/mqtt_wrapper.py
```diff
@@ -116,6 +123,13 @@ def _on_connect(self, client, userdata, flags, rc):
         if self.on_connect_cb is not None:
             self.on_connect_cb()
 
+    def _on_disconnect(self, userdata, rc):
+        if rc != 0:
+            self.logger.error(
+                "MQTT unexpected disconnection (network or broker originated)"
```
GwendalRaoul, line 129: Should  ...%s (%s)", connack_string(rc), rc) be added like in _on_connect.
Better to have all info available

python_transport/wirepas_gateway/protocol/mqtt_wrapper.py
```diff
@@ -241,6 +261,7 @@ def run(self):
             # As this thread is daemonized, inform the parent that this
             # thread has exited
             self.on_termination_cb()
+            exit(2)
```
GwendalRaoul, line 264: There is no semantic at the moment on exit code but at least it is not homogeneous. Above we call exit(-1), and here exit(2). So either both negative or both positive.

GwendalRaoul: But does it make sense to call exit() here?
termination_cb is already exiting after stopping the dbus client.

So it would be more clean to exit from transport_service instead.
But the other exit in this file should be removed too, so I'll move all of them at the same time later

ansinnovation: I will remove it. I guess it's a workaround for the dbus client issue, which is fixed now.

python_transport/wirepas_gateway/protocol/mqtt_wrapper.py
```diff
@@ -171,19 +190,21 @@ def _get_socket(self):
         if sock is not None:
             return sock
 
-        self.logger.error("MQTT, unexpected disconnection")
-
-        if not self.connected:
+        if self.connected:
```
GwendalRaoul, line 193: If it is only needed here, I guess self._client.is_connected() should be enough. No need to duplicate the mqtt connection status here.

ansinnovation: It seems that is_connected() is available since paho 1.5.0, and we are still using 1.4.0

GwendalRaoul: Good point! I probably selected wrong version when checking ;-)

GwendalRaoul (approved):

PR #224, opened 2021-06-09
[RFC] add buffering monitoring support

ansinnovation:
Here is the kind of messages that are generated:
(With blackhole detection)
Net-State: MQTT CONNECTED, MQTT messages sent 1, messages received 0
           MQTT buffers 0 (1000 max), publish delay 0 sec (10 max), sinks cost LOW
           Sink [sink1]  messages received 0, last message 0 s ago
           Sink [sink2]  messages received 0, last message 0 s ago

(Without blackhole detection)
Net-State: MQTT CONNECTED, MQTT messages sent 1, messages received 0
           MQTT buffers 0, publish delay 0 sec
           Sink [sink1]  messages received 0, last message 0 s ago
           Sink [sink2]  messages received 0, last message 0 s ago

### DIFF
--- a/python_transport/wirepas_gateway/transport_service.py
+++ b/python_transport/wirepas_gateway/transport_service.py
@@ -7,7 +7,7 @@ import os
 import wirepas_mesh_messaging as wmm
 from time import time, sleep
 from uuid import getnode
-from threading import Thread
+from threading import Thread, Lock
 
 from wirepas_gateway.dbus.dbus_client import BusClient
 from wirepas_gateway.protocol.topic_helper import TopicGenerator, TopicParser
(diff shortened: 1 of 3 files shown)


python_transport/wirepas_gateway/transport_service.py
```diff
@@ -238,6 +324,15 @@ def __init__(self, settings, logger=None, **kwargs):
         else:
             self.data_event_id = None
 
+        self.net_monitoring_thread = NetMonitoringThread(
```
GwendalRaoul, line 327: Should be started only if WM_GW_BUFFERING_MONITOR_PERIOD is different than 0

ansinnovation: I would say that we have two options:

1) We keep the same pattern as the monitoring_thread: we avoid the object creation if not needed, and we check if the object is not None at each call site.
The transport_service has a dependancy on how the configuration is done, but we save some memory.

2) Or we always create the object. Then avoiding to start the Thread makes little sense because we are still adding a dependancy on how the configuration is done, but for no measurable gain.


I'm ok with both. But I don't see the in-between solution as an improvment.
Let me know.

python_transport/wirepas_gateway/utils/argument_tools.py
```diff
@@ -402,6 +402,17 @@ def add_buffering_settings(self):
             ),
         )
 
+        self.buffering.add_argument(
+            "--buffering_monitor_period",
+            default=os.environ.get("WM_GW_BUFFERING_MONITOR_PERIOD", 0),
```
ansinnovation, line 407: Should we enable the feature by default ?
I would say that a default set to one message every 6H could be helpful for debugging.

PR #229, opened 2022-03-11
Gateway status: monitor it from gateway

GwendalRaoul:
The gateway discovery mechanism is based on retain topic on
gw-event/status/<gw-id>

The gateway publishes its online status as retain message when connection
is established and the last will message set it to offline.

It was seen sometimes that this retain status was removed at broker level,
either because of broker error or because it was erased manually.
In order to mitigate this issue, try to detect it from gateway itself.

It will work only if gateway can subscribe to its own event-topic


kboulaid (approved):
This enhancement is welcomed.

PR #236, opened 2022-10-25
Gateway status

LePailleurThibault:
Updating the transport service to handle request on the gateway status

Adding Gateways status request and response topics.
This can be used in the case where the mqtt broker does not support retained message
and we would like to know the status of the gateways as soon as possible.

The concerned topics in the mqtt broker are :
gw-request/get_gw_status
gw-response/get_gw_status/<gw-id>/<sink-id>

### DIFF
--- a/python_transport/wirepas_gateway/protocol/mqtt_wrapper.py
+++ b/python_transport/wirepas_gateway/protocol/mqtt_wrapper.py
@@ -44,6 +44,10 @@ class MQTTWrapper(Thread):
         # Keep track of latest published packet
         self._publish_monitor = PublishMonitor()
 
+        # load special settings for borker compatibility
+        self.max_qos = settings.mqtt_max_qos_supported
+        self.retain_supported = settings.mqtt_retain_flag_supported
+
         if settings.mqtt_use_websocket:
             transport = "websockets"
             self._use_websockets = True
@@ -234,7 +238,7 @@ class MQTTWrapper(Thread):
 
     def _set_last_will(self, topic, data):
         # Set Last wil message
-        self._client.will_set(topic, data, qos=2)
+        self._client.will_set(topic, data, qos=self.max_qos, retain=self.retain_supported)
 
     def run(self):
         self.running = True
@@ -279,6 +283,13 @@ class MQTTWrapper(Thread):
             retain: Is it a retain message
 
         """
+        # Limit qos in case broker has a limit
+        if qos > self.max_qos:
+            qos = self.max_qos
+
+        # Clear retain flag if not supported
+        retain = retain and self.retain_supported
+
         mid = self._client.publish(topic, payload, qos=qos, retain=retain).mid
         self._unpublished_mid_set.add(mid)
 
@@ -288,15 +299,33 @@ class MQTTWrapper(Thread):
         Args:
             topic: Topic to publish on
             payload: Payload
-            qos: Qos to use
-            retain: Is it a retain message
+            qos: Qos to use. Can be less than requested if broker does
+                not support it
+            retain: Is it a retain message. Can be discarded if broker
+                    does not support it
 
         """
+        # No need to check qos or retain at this stage as
+        # done later in real publish to broker
+
         # Send it to the queue to be published from Mqtt thread
         self._publish_queue.put((topic, payload, qos, retain))
         self._publish_monitor.on_publish_request()
 
     def subscribe(self, topic, cb, qos=2) -> None:
+        """ Method to subscribe to mqtt topic
+
+        Args:
+            topic: Topic to publish on
+            cb: Callback to call on message reception
+            qos: Qos to use. Can be less than requested if broker does
+                 not support it
+
+        """
+        # Limit qos in case broker has a limit
+        if qos > self.max_qos:
+            qos = self.max_qos
+
         self.logger.debug("Subscribing to: {}".format(topic))
         self._client.subscribe(topic, qos)
         self._client.message_callback_add(topic, cb)
--- a/python_transport/wirepas_gateway/protocol/mqtt_wrapper.py
+++ b/python_transport/wirepas_gateway/protocol/mqtt_wrapper.py
@@ -234,7 +234,7 @@ class MQTTWrapper(Thread):
 
     def _set_last_will(self, topic, data):
         # Set Last wil message
-        self._client.will_set(topic, data, qos=2, retain=True)
+        self._client.will_set(topic, data, qos=2)
 
     def run(self):
         self.running = True
--- a/python_transport/wirepas_gateway/transport_service.py
+++ b/python_transport/wirepas_gateway/transport_service.py
@@ -275,8 +275,6 @@ class TransportService(BusClient):
         self.mqtt_wrapper.start()
 
         self.logger.info("Gateway started with id: %s", self.gw_id)
-        self.logger.info(f"wmm version: {wmm.__version__}")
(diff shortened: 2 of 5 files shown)


python_transport/wirepas_gateway/protocol/mqtt_wrapper.py
```diff
@@ -288,15 +299,33 @@ def publish(self, topic, payload, qos=1, retain=False) -> None:
...
+
+        Args:
+            topic: Topic to publish on
+            cb: Callback to call on message reception
+            qos: Qos to use. Can be less than requested if broker does
+                 not support it
+
+        """
+        # Limit qos in case broker has a limit
+        if qos > self.max_qos:
```
pgaucherand, line 326: same comment as the publish method

python_transport/wirepas_gateway/protocol/mqtt_wrapper.py
```diff
@@ -288,15 +299,33 @@ def publish(self, topic, payload, qos=1, retain=False) -> None:
...
 
         """
+        # No need to check qos or retain at this stage as
+        # done later in real publish to broker
+
         # Send it to the queue to be published from Mqtt thread
         self._publish_queue.put((topic, payload, qos, retain))
         self._publish_monitor.on_publish_request()
 
     def subscribe(self, topic, cb, qos=2) -> None:
```
pgaucherand, line 315: General comment around Qos1. Although that is probably something to handle in the callbacks rather than in this part of the code. Should we think of a mechanism to handle duplicates? I am thinking specifically about instructions coming in that would result in WM duplicate instructions. For example, it might be possible to receive to sendData to WM stack.

python_transport/wirepas_gateway/utils/argument_tools.py
```diff
@@ -363,6 +363,26 @@ def add_mqtt(self):
...
+            help=("Set to true if broker support retain flag"),
+        )
+
+        self.mqtt.add_argument(
+            "--mqtt_max_qos_supported",
+            default=os.environ.get("WM_SERVICES_MQTT_MAX_QOS_SUPPORTED", 2),
+            action="store",
+            type=self.str2int,
+            help=(
+                "Max qos supported by broker in case it has limitation"
```
pgaucherand, line 381: I would suggest something like: "Max QOS supported by the broker. This allows for automatically downgrading the QOS of a published message or a message received on a given topic."

python_transport/wirepas_gateway/transport_service.py
```diff
@@ -159,6 +159,68 @@ def initialize_sink(self, name):
                 sink.cost = self.minimum_sink_cost
 
 
+class SendGatewayStatusThread(Thread):
+
+    def __init__(
+        self,
+        period,
```
pgaucherand, line 166: should we have a default parameter for the period?

python_transport/wirepas_gateway/transport_service.py
```diff
@@ -173,6 +235,9 @@ class TransportService(BusClient):
     # Period in s to check for black hole issue
     MONITORING_BUFFERING_PERIOD_S = 1
 
+    # Period in s to send gateway status to the backend
+    SEND_GATEWAY_STATUS_PERIOD_S = 20
```
pgaucherand, line 239: see comment above for a default value.

pgaucherand: perhaps we would also want to configure this from environment variable (i..e add a parameter to argument parser tool.

pgaucherand (changes requested):
As discussed, I wonder if we need to look at the implications of dealing with duplicate messages when using Qos1 especially so on subscribed topics so that we don't generate duplicate load on the mesh network.

python_transport/wirepas_gateway/protocol/mqtt_wrapper.py
```diff
@@ -279,6 +283,13 @@ def _publish_from_wrapper_thread(self, topic, payload, qos, retain):
             retain: Is it a retain message
 
         """
+        # Limit qos in case broker has a limit
+        if qos > self.max_qos:
+            qos = self.max_qos
```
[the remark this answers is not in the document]
GwendalRaoul: And in general, max qos should be handled in a dedicated PR.

Let's address only retain here.

python_transport/wirepas_gateway/protocol/mqtt_wrapper.py
GwendalRaoul: yes definitely something that could happen.

python_transport/wirepas_gateway/protocol/mqtt_wrapper.py
```diff
@@ -288,15 +299,33 @@ def publish(self, topic, payload, qos=1, retain=False) -> None:
         Args:
             topic: Topic to publish on
             payload: Payload
-            qos: Qos to use
-            retain: Is it a retain message
+            qos: Qos to use. Can be less than requested if broker does
```
GwendalRaoul, line 302: I think in general, we should just prevent usage the qos2 in all the code and have our own duplicate detection mechanism.
As far as I know, all broker support qos1

python_transport/wirepas_gateway/transport_service.py
```diff
@@ -159,6 +159,68 @@ def initialize_sink(self, name):
...
+
+        self.gw_id = gw_id
+        self.running = False
+
+    def publish_status(self):
+        """
+        Publish the gateway status in the MQTT Broker.
+        """
+        event_online = wmm.StatusEvent(self.gw_id, wmm.GatewayState.ONLINE)
+        status_topic = TopicGenerator.make_status_topic(self.gw_id)
```
GwendalRaoul, line 195: small optimization: status_topic could be generated only once in constructor.

python_transport/wirepas_gateway/protocol/mqtt_wrapper.py
```diff
@@ -44,6 +44,10 @@ def __init__(
         # Keep track of latest published packet
         self._publish_monitor = PublishMonitor()
 
+        # load special settings for borker compatibility
```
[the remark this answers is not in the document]
teemupiiroinenwirepas: All the other comments seem to start with capital letter

python_transport/wirepas_gateway/transport_service.py
teemupiiroinenwirepas: period should have unit (s/ms)?

python_transport/wirepas_gateway/transport_service.py
```diff
@@ -159,6 +159,68 @@ def initialize_sink(self, name):
...
+        period,
+        mqtt_wrapper,
+        gw_id,
+    ):
+        """
+        Thread sending periodically Gateway Status to the MQTT broker.
+
+        Args:
+            period: the period to check the buffer status
+            mqtt_wrapper: the mqtt wrapper to get access to queue level
```
teemupiiroinenwirepas, line 175: Is the comment true? Connection state is checked from this object too.

python_transport/wirepas_gateway/transport_service.py
LePailleurThibault: Yes, I'm going to add it explicitly

LePailleurThibault:
This branch has been splitted into 2 other branches: 
    - retain_control (for the new retain option),
    - qos_mqtt_compatibility (for the implementation of the cache and the new gestion of QoS inside the MQTT).

PR #237, opened 2022-10-27
Adding a retain flag supported by mqtt in the gateway and status topi…

LePailleurThibault:
…cs in mqtt.

Some brokers (like AWS iot core) do not support retain messages. Add option to transport to specify if the retain is supported by broker.

New MQTT topics can be used in the case where the mqtt broker does not support retained message and we would like to know the status of the gateways as soon as possible.

The concerned topics in the mqtt broker are :
gw-request/get_gw_status
gw-response/get_gw_status/<gw-id>/<sink-id>

Closes # .


pgaucherand (commented):
Just a minor comment from, otherwise looks good.

PR #238, opened 2022-11-04
Qos mqtt compatibility

LePailleurThibault:
MQTT compatibility with QoS

Some MQTT brokers do not support qos=2.

Changing transport service to publish with qos=1.
Adding a cache in transport service to avoid sending redundant messages.

### DIFF
--- a/python_transport/wirepas_gateway/protocol/cache_message.py
+++ b/python_transport/wirepas_gateway/protocol/cache_message.py
@@ -1,0 +1,76 @@
+# Copyright 2019 Wirepas Ltd licensed under Apache License, Version 2.0
+#
+# See file LICENSE for full license details.
+#
+from time import time
+from threading import Lock
+
+
+class CacheMessage:
+    """
+    Class to cache messages to avoid sending duplicates in MQTT with QoS=1.
+    """
+    def __init__(
+        self,
+        cache_time_window,
+        cache_update_s
+    ):
+        """
+        Args:
+            cache_time_window: time in seconds after which a message is clean from the cache.
+            cache_update_s: period in seconds to update the list of received messages.
+                            Must be shorter than cache_time_window
+        """
+        self._lock = Lock()
+        self.msg_list = dict()  # Dictionary of received messages id mapped to their timestamps
+
+        self.cache_time_window = cache_time_window
+        self.cache_update_s = min(cache_update_s, cache_time_window)
+
+        self.last_update_time = 0  # last time an update of the list of received messages was done
+
+    def add_msg(self, msg_id):
+        """
+        Adds a received message to the cache.
+        If the message is a duplicate, it only updates its last timestamp to the current time.
+
+        Args:
+            msg_id: the id of a message to be added to the cache.
+        """
+        current_time = time()
+        with self._lock:
+            if current_time - self.last_update_time > self.cache_update_s:
+                self._clean_msg_list()
+
+            is_duplicate = self.is_in_cache(msg_id)
+            self.msg_list[msg_id] = current_time
+            return not is_duplicate
+
+    def get_size(self):
+        """
+        Returns the number of messages in the cache.
+        """
+        with self._lock:
+            return len(self.msg_list)
+
+    def _clean_msg_list(self):
+        """
+        Clean the old messages in the cache
+        based on the time window cache_time_window attribute.
+        """
+        current_time = time()
+        self.msg_list = {
+            msg_id: msg_time for (msg_id, msg_time) in self.msg_list.items()
+            if current_time - msg_time <= self.cache_time_window
+        }
+        self.last_update_time = current_time
+
+    def is_in_cache(self, msg_id):
+        """
+        Returns True if the id of a message is already in the cache.
+        Return False otherwise.
+
+        Args:
+            msg_id: id of the message received
+        """
+        return msg_id in self.msg_list
--- a/python_transport/wirepas_gateway/protocol/mqtt_wrapper.py
+++ b/python_transport/wirepas_gateway/protocol/mqtt_wrapper.py
@@ -234,7 +234,7 @@ class MQTTWrapper(Thread):
 
     def _set_last_will(self, topic, data):
         # Set Last wil message
-        self._client.will_set(topic, data, qos=2, retain=True)
+        self._client.will_set(topic, data, qos=1, retain=True)
 
     def run(self):
         self.running = True
@@ -296,7 +296,16 @@ class MQTTWrapper(Thread):
         self._publish_queue.put((topic, payload, qos, retain))
         self._publish_monitor.on_publish_request()
 
-    def subscribe(self, topic, cb, qos=2) -> None:
+    def subscribe(self, topic, cb, qos=1) -> None:
+        """ Method to subscribe to mqtt topic
+
+        Args:
+            topic: Topic to subscribe to
+            cb: Callback to call on message reception
+            qos: Qos to use. Can be less than requested if broker does
+                 not support it
+
+        """
         self.logger.debug("Subscribing to: {}".format(topic))
         self._client.subscribe(topic, qos)
         self._client.message_callback_add(topic, cb)
--- a/python_transport/wirepas_gateway/transport_service.py
+++ b/python_transport/wirepas_gateway/transport_service.py
@@ -7,7 +7,7 @@ import os
 import wirepas_mesh_messaging as wmm
 from time import time, sleep
 from uuid import getnode
-from threading import Thread
+from threading import Thread, Lock
 
 from wirepas_gateway.dbus.dbus_client import BusClient
 from wirepas_gateway.protocol.topic_helper import TopicGenerator, TopicParser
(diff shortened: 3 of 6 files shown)


python_transport/wirepas_gateway/protocol/cache_message.py
```diff
@@ -0,0 +1,76 @@
+# Copyright 2019 Wirepas Ltd licensed under Apache License, Version 2.0
+#
+# See file LICENSE for full license details.
+#
+from time import time
+from threading import Lock
+
+
+class CacheMessage:
```
pgaucherand, line 9: I would rename it to MessageCache

python_transport/wirepas_gateway/protocol/cache_message.py
```diff
@@ -0,0 +1,76 @@
...
+    Class to cache messages to avoid sending duplicates in MQTT with QoS=1.
+    """
+    def __init__(
+        self,
+        cache_time_window,
+        cache_update_s
+    ):
+        """
+        Args:
+            cache_time_window: time in seconds after which a message is clean from the cache.
```
pgaucherand, line 20: is clean from the cache => is removed from the cache

python_transport/wirepas_gateway/transport_service.py
```diff
@@ -234,6 +235,8 @@ def __init__(self, settings, logger=None, **kwargs):
         else:
             self.data_event_id = None
 
+        self.msg_manager = CacheMessage(settings.cache_time_window, settings.cache_update_s)
```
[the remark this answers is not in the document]
pgaucherand: the naming should be consistent. Either rename MessageCache to MessageManager or rename self.msg_manager to self.message_cache

python_transport/wirepas_gateway/transport_service.py
```diff
@@ -458,6 +458,9 @@ def _on_send_data_cmd_received(self, client, userdata, message):
             self.logger.error(str(e))
             return
 
+        if not self.msg_manager.add_msg(request.req_id):
```
pgaucherand, line 461: a comment here would be welcome to add a brief explanation around the logic at play

python_transport/wirepas_gateway/transport_service.py
```diff
@@ -499,6 +502,9 @@ def _on_get_configs_cmd_received(self, client, userdata, message):
...
```
pgaucherand, line 505: same comment as before

python_transport/wirepas_gateway/transport_service.py
```diff
@@ -543,6 +549,9 @@ def _on_get_gateway_info_cmd_received(self, client, userdata, message):
...
```
pgaucherand, line 552: same comment as before

python_transport/wirepas_gateway/transport_service.py
```diff
@@ -566,6 +575,9 @@ def _on_set_config_cmd_received(self, client, userdata, message):
...
```
pgaucherand, line 578: same comment as before

python_transport/wirepas_gateway/transport_service.py
```diff
@@ -594,6 +606,9 @@ def _on_otap_status_request_received(self, client, userdata, message):
...
```
pgaucherand, line 609: same comment as before

python_transport/wirepas_gateway/transport_service.py
```diff
@@ -646,6 +661,9 @@ def _on_otap_upload_scratchpad_request_received(self, client, userdata, message)
 
         self.logger.info("OTAP upload request received for %s", request.sink_id)
 
+        if not self.msg_manager.add_msg(request.req_id):
```
pgaucherand, line 664: same comment as before

python_transport/wirepas_gateway/transport_service.py
```diff
@@ -672,6 +690,9 @@ def _on_otap_process_scratchpad_request_received(self, client, userdata, message
             self.logger.error(str(e))
             return
 
+        if not self.msg_manager.add_msg(request.req_id):
```
pgaucherand, line 693: same comment as before

python_transport/wirepas_gateway/transport_service.py
```diff
@@ -707,6 +728,9 @@ def _on_otap_set_target_scratchpad_request_received(
             self.logger.error("Action is mandatory")
             res = wmm.GatewayResultCode.GW_RES_INVALID_PARAM
 
+        if not self.msg_manager.add_msg(request.req_id):
```
pgaucherand, line 731: same comment as before

python_transport/wirepas_gateway/utils/argument_tools.py
```diff
@@ -402,6 +402,32 @@ def add_buffering_settings(self):
             ),
         )
 
+    def add_cache_settings(self):
+        """ Parameters used to avoid sending redundant messages in the mqtt broker """
+        cleaning_time_window_default = 20*60  # 20 minutes
+        cache_update_s_default = 20  # 20 secondes
```
pgaucherand, line 408: should move to line 419 or thereabouts

pgaucherand (changes requested):
Some minor changes with naming and comments to make but otherwise looks good.

python_transport/wirepas_gateway/protocol/mqtt_wrapper.py
```diff
@@ -296,7 +296,16 @@ def publish(self, topic, payload, qos=1, retain=False) -> None:
         self._publish_queue.put((topic, payload, qos, retain))
         self._publish_monitor.on_publish_request()
 
-    def subscribe(self, topic, cb, qos=2) -> None:
+    def subscribe(self, topic, cb, qos=1) -> None:
```
GwendalRaoul, line 299: I would remove the qos option everywhere.
qos=2 cannot be used anymore and qos=0, we are not using it. So I would just use qos=1

python_transport/wirepas_gateway/protocol/cache_message.py
```diff
@@ -0,0 +1,76 @@
+# Copyright 2019 Wirepas Ltd licensed under Apache License, Version 2.0
```
GwendalRaoul, line 1: 2022

python_transport/wirepas_gateway/protocol/cache_message.py
```diff
@@ -0,0 +1,76 @@
...
+        cache_update_s
+    ):
+        """
+        Args:
+            cache_time_window: time in seconds after which a message is clean from the cache.
+            cache_update_s: period in seconds to update the list of received messages.
+                            Must be shorter than cache_time_window
+        """
+        self._lock = Lock()
+        self.msg_list = dict()  # Dictionary of received messages id mapped to their timestamps
```
GwendalRaoul, line 25: self._msg_list? to keep same convention for internal variable

python_transport/wirepas_gateway/protocol/cache_message.py
```diff
@@ -0,0 +1,76 @@
...
+        """
+        Adds a received message to the cache.
+        If the message is a duplicate, it only updates its last timestamp to the current time.
+
+        Args:
+            msg_id: the id of a message to be added to the cache.
+        """
+        current_time = time()
+        with self._lock:
+            if current_time - self.last_update_time > self.cache_update_s:
```
GwendalRaoul, line 42: The cache_update_s is a bit misleading as at the end it will be only evaluated when a message is received.

I think it would be cleaner to do it asynchronously in a dedicated thread than synchronously with an addition

GwendalRaoul (changes requested):

python_transport/wirepas_gateway/protocol/cache_message.py
```diff
@@ -0,0 +1,76 @@
...
+        based on the time window cache_time_window attribute.
+        """
+        current_time = time()
+        self.msg_list = {
+            msg_id: msg_time for (msg_id, msg_time) in self.msg_list.items()
+            if current_time - msg_time <= self.cache_time_window
+        }
+        self.last_update_time = current_time
+
+    def is_in_cache(self, msg_id):
```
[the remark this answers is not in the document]
teemupiiroinenwirepas: Is there some time stamp in the incoming message? In some cases you might send broadcast RemoteAPI message that is exactly the same as previous message after one or two minutes (or even faster in LL network) to be sure that the message goes through. In this case the hash would match also.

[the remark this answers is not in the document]
GwendalRaoul: The uniqueness of request_id is specified here: https://github.com/wirepas/backend-apis/blob/master/gateway_to_backend/README.md#request-unique-identifier
So it is written that at least 48 bits must be unique.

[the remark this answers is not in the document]
teemupiiroinenwirepas: Good. Is the some check in gateway for req id? If I remember correctly, at least earlier, it was possible to use any constant req id (e.g. always zero).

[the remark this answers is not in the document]
GwendalRaoul: At the moment there is no check, it is just copied to the response. So if you use same id, you cannot match the response to an answer.
But with this proposed changed, this id wooul now be used as a way to detect duplicate messages. We could change it as a hash of full message, but will still filter twos messages with same id and same content sent very close in time.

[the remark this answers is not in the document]
teemupiiroinenwirepas: In our case we are not so interested about the id as we cannot rely that the message reaches the target, so we need to resend it until we get response from the node.

PR #242, opened 2022-12-08
Rtc service

LePailleurThibault:
Implementation of the RTC service on top of DBUS

RTC service sends periodically time to the network.
It also gets the expected RTC time from the network to display the time difference with its own value during log messages
Add of Docker file to create the container using RTC service

### DIFF
--- a/python_transport/setup.py
+++ b/python_transport/setup.py
@@ -106,6 +106,7 @@ setup(
             "wm-gw=wirepas_gateway.transport_service:main",
             "wm-dbus-print=wirepas_gateway.dbus_print_client:main",
             "wm-node-conf=wirepas_gateway.configure_node:main",
+            "wm-rtc=wirepas_gateway.rtc_service:main",
         ]
     },
 )
--- a/python_transport/wirepas_gateway/rtc_service.py
+++ b/python_transport/wirepas_gateway/rtc_service.py
@@ -1,0 +1,375 @@
+# Copyright 2022 Wirepas Ltd licensed under Apache License, Version 2.0
+#
+# See file LICENSE for full license details.
+#
+import datetime
+import logging
+import os
+import struct
+import sys
+from time import sleep, time
+from uuid import getnode
+from threading import Thread
+
+from wirepas_gateway.dbus.dbus_client import BusClient
+from wirepas_gateway.utils import ParserHelper
+
+from wirepas_gateway import __version__ as transport_version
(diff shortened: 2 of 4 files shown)


python_transport/setup.py
```diff
@@ -106,6 +106,7 @@ def get_systemd_lib(key):
             "wm-gw=wirepas_gateway.transport_service:main",
             "wm-dbus-print=wirepas_gateway.dbus_print_client:main",
             "wm-node-conf=wirepas_gateway.configure_node:main",
+            "wm-rtc=wirepas_gateway.rtc_service:main",
```
GwendalRaoul, line 109: Nope, it shouldn't be package with the transport wheel for now.
We should create a new wheel for BusClient and then different one for the other services

python_transport/wirepas_gateway/rtc_service.py
```diff
@@ -0,0 +1,375 @@
...
+from threading import Thread
+
+from wirepas_gateway.dbus.dbus_client import BusClient
+from wirepas_gateway.utils import ParserHelper
+
+from wirepas_gateway import __version__ as transport_version
+from wirepas_gateway import __pkg_name__
+
+# This constant is the actual API level implemented by this rtc module (cf WP-RM-128)
+IMPLEMENTED_API_VERSION = 2
```
GwendalRaoul, line 21: Copy/paste??? Why keeping it here?

python_transport/wirepas_gateway/utils/argument_tools.py
```diff
@@ -541,6 +541,33 @@ def add_filtering_config(self):
             ),
         )
 
+    def add_rtc_settings(self):
```
GwendalRaoul, line 544: Your new service should have its own arugment parser that has nothing to do with transport one

python_transport/wirepas_gateway/rtc_service.py
```diff
@@ -0,0 +1,375 @@
...
+        description="Wirepas Gateway Transport service arguments",
+        version=transport_version,
+    )
+
+    parse.add_file_settings()
+    parse.add_gateway_config()
+    parse.add_filtering_config()
+    parse.add_buffering_settings()
+    parse.add_deprecated_args()
+    parse.add_rtc_settings()
```
GwendalRaoul, line 337: This must not be in the rtc service

GwendalRaoul (changes requested):

PR #243, opened 2023-01-03
Rtc service

LePailleurThibault:
Implementation of the RTC service sending periodically time to the network.

### DIFF
--- a/docker/rtc_service/Dockerfile
+++ b/docker/rtc_service/Dockerfile
@@ -1,0 +1,64 @@
+FROM python:3.10.8-alpine3.17 AS builder
+
+RUN adduser --disabled-password wirepas
+
+RUN apk add --no-cache \
+		gcc \
+		bash \
+		build-base \
+		make \
+		musl-dev \
+		elogind-dev \
+		python3-dev \
+		py3-gobject3 \
+		cairo-dev \
+		gobject-introspection-dev
+
+RUN python3 -m pip install wheel setuptools pkgconfig
+
+USER wirepas
+WORKDIR /home/wirepas
+
+COPY --chown=wirepas ./python_transport /home/wirepas/python_transport
+WORKDIR /home/wirepas/python_transport
+
+RUN ./utils/generate_wheel.sh
+
+USER wirepas
+
+# Note:
+# We are re-using the transport service module builder for the dbus client.
+# This is a hack as dbus client is part of python_transport.
+# But, they should be splitted later and the dbus client wheel
+# will be independent from the transport service.
+
+RUN pip3 install dist/wirepas_gateway*.whl --no-deps --user
+# Dependencies are installed manually as runner image already have wmm
+# Todo: removing wmm from requirement list would be better
+RUN pip3 install pydbus==0.6.0 PyYAML==6.0.1 --user
+RUN pip3 install gobject PyGObject --user
+RUN pip3 install ntplib --user
+
+
+# Build the final image with prebuilt wmm image
+FROM wirepas/wmm_alpine_cpp:1.2.5
+
+USER root
+
+# Variable set from CI
+ARG GATEWAY_BUILD_SHA1=unset
+
+RUN apk add --no-cache libelogind glib gobject-introspection
+
+USER wirepas
+
+ENV PATH="/home/wirepas/.local/bin:${PATH}"
+
+# Copy the built wheel and its dependencies from builder
+COPY --from=builder /home/wirepas/.local /home/wirepas/.local
+
+COPY ./rtc_service/*.py /home/wirepas/rtc_service/
+
+CMD ["python3", "/home/wirepas/rtc_service/rtc_service.py"]
+
+LABEL com.wirepas.gateway.build.sha1="${GATEWAY_BUILD_SHA1}"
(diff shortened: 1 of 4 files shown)


docker/rtc_service/Dockerfile
```diff
@@ -0,0 +1,63 @@
+FROM python:3.8.8-alpine3.12 AS builder
```
GwendalRaoul, line 1: It has to be changed to 3.17 now as it is the base image for other services

docker/rtc_service/Dockerfile
```diff
@@ -0,0 +1,63 @@
...
+USER wirepas
+WORKDIR /home/wirepas
+
+COPY --chown=wirepas ./python_transport /home/wirepas/python_transport
+WORKDIR /home/wirepas/python_transport
+
+RUN ./utils/generate_wheel.sh
+
+USER wirepas
+
```
GwendalRaoul, line 28: Add a note here that this is a hack as dbus client is part of python_transport, but later they should be splitted

rtc_service/rtc_service.py
```diff
@@ -0,0 +1,288 @@
...
+        self.synchronization_thread.start()
+
+    def on_stack_started(self, name):
+        logging.debug("Sink started: %s", name)
+
+    def on_stack_stopped(self, name):
+        logging.debug("Sink stopped: %s", name)
+
+    def on_sink_connected(self, name):
+        logging.info("Sink connected, sending new configs")
```
GwendalRaoul, line 183: Is the message true ?
You say, "sending new configs" but nothing is sent!

rtc_service/rtc_service.py
```diff
@@ -0,0 +1,288 @@
...
+        self.synchronization_thread = SynchronizationThread(
+            settings.rtc_synchronization_period_s,
+            settings.timezone_offset_s,
+            settings.timezone_from_gateway_clock,
+            settings.get_time_from_local,
+            self.sink_manager,
+        )
+        self.synchronization_thread.start()
+
+    def on_stack_started(self, name):
```
GwendalRaoul, line 176: Is it really needed to override those calls for just a debug message?

rtc_service/rtc_service.py
```diff
@@ -0,0 +1,288 @@
...
+        Main service for rtc service module
+    """
+    parser = argparse.ArgumentParser(fromfile_prefix_chars='@')
+
+    parser.add_argument(
+        "--rtc_synchronization_period_s",
+        default=os.environ.get("WM_RTC_SYNCHRONIZATION_PERIOD_S", 20*60),
+        action="store",
+        type=str2int,
+        help=("Period of time before sending a new rtc time in the network."),
```
GwendalRaoul, line 224: before sending -> to send

rtc_service/rtc_service.py
```diff
@@ -0,0 +1,288 @@
...
+              "True means that the time is taken directly from gateway."
+              "Note: You must assure that gateway are synchronized if set to True"),
+    )
+
+    settings = parser.parse_args()
+
+    # Set default debug level
+    debug_level = "info"
+    try:
+        debug_level = os.environ["DEBUG_LEVEL"]
```
GwendalRaoul, line 261: This is a new service so you don't have to keep backward compatibility

rtc_service/rtc_service.py
```diff
@@ -0,0 +1,288 @@
...
+        """
+        Main loop that send periodically gateway time to the network.
+        """
+        self.running = True
+
+        while self.running:
+            self.publish_time()
+            sleep(self.period)
+
+    def stop(self):
```
GwendalRaoul, line 152: It looks like you never call stop.
Is it needed to keep it?

Thread is daemonized so should be killed when the process is closed

GwendalRaoul (changes requested):

rtc_service/rtc_service.py
LePailleurThibault: I'm going to remove this function. 
stop() function in the transport service is also never called and it's also daemonized. Should we remove this one as well ?

rtc_service/rtc_service.py
LePailleurThibault: Logs are useful but, you are right. If the stack starts, at least the transport service should logs this message. We don't need to duplicate the logs.

rtc_service/rtc_service.py
```diff
@@ -0,0 +1,263 @@
...
+                hop_limit=0,
+            )
+            logging.debug("time elapsed to send RTC time through the gateway: "
+                          f"{int((time()-start)*1000)}ms")
+
+    def run(self):
+        """
+        Main loop that send periodically gateway time to the network.
+        """
+        self.running = True
```
GwendalRaoul, line 147: self.running is useless

rtc_service/rtc_service.py
```diff
@@ -0,0 +1,263 @@
...
+        data_payload = RTC_VERSION.to_bytes(2, "little") + encode_tlv(timestamp_ms, self.timezone_offset_s)
+        logging.info("Send rtc message to the network")
+        for sink in self.sink_manager.get_sinks():
+            start = time()
+            sink.send_data(
+                dst=BROADCAST_ADDRESS,
+                src_ep=RTC_SOURCE_EP,
+                dst_ep=RTC_DEST_EP,
+                qos=0,
+                initial_time=INITIAL_DELAY_MS,
```
GwendalRaoul, line 135: you can remove this line, it is deprecated feature

GwendalRaoul (approved):

PR #249, opened 2023-03-14
Transport: fill and update the extended status when needed

GwendalRaoul:
It will be merged only when official wheel is available

Closes # .

### DIFF
--- a/python_transport/wirepas_gateway/transport_service.py
+++ b/python_transport/wirepas_gateway/transport_service.py
@@ -8,7 +8,7 @@ import sys
 import wirepas_mesh_messaging as wmm
 from time import time, sleep
 from uuid import getnode
-from threading import Thread
+from threading import Lock, Thread
 
 from wirepas_gateway.dbus.dbus_client import BusClient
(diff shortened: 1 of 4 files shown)


python_transport/wirepas_gateway/transport_service.py
```diff
@@ -237,12 +243,62 @@ def _on_mqtt_wrapper_termination_cb(self):
...
+
+        def set_status_after_delay(delay_s):
+            sleep(delay_s)
+            logging.debug("Time to update the status")
+            self._is_update_status_scheduled = False
+            self._set_status()
+
+        # We are here as there was no update scheduled yet.
+        # Wait a bit
+        thread = Thread(target=set_status_after_delay, args=[0.5])
```
LePailleurThibault, line 287: Out of curiosity, I know the function is fast to execute and it doesn't change a lot but why didn't you daemonized the thread ?

LePailleurThibault (approved):
That's a good for me :)
We just need to check the docker containers before merging

PR #264, opened 2024-03-13
RTC service force retry sending RTC time

LePailleurThibault:
Retry to get the NTP time more frequently when a connection issue happened previously trying to connect to a NTP server.
Catch socket errors that could occur when trying to get the time from a NTP server without a stable connection.

### DIFF
--- a/rtc_service/rtc_service.py
+++ b/rtc_service/rtc_service.py
@@ -7,6 +7,7 @@ import datetime
 import logging
 import ntplib
 import os
+import socket
 import struct
 import sys
 from time import sleep, time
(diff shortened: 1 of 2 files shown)


rtc_service/rtc_service.py
```diff
@@ -141,7 +150,10 @@ def run(self):
         """
         while True:
             self.publish_time()
```
GwendalRaoul, line 152: Why using a global variable and not returning True or False from publish_time ?
It seems a bit cleaner to me.
```python
if  self.publish_time():
        sleep(self.period)
else:
        sleep(self.retry_period)
```

LePailleurThibault: You are right, I will correct it

GwendalRaoul (approved):

PR #276, opened 2024-09-06
Enable clearing of sink scratchpad

sgunes-wirepas:
According to the gateway to backend API, scratchpad should be cleared if the scratchpad bytes are not present in the upload scratchpad request.

In order to make it easier to test OTAP functionality of the gateway, this change makes it possible to clear the scratchpad.

### DIFF
--- a/python_transport/wirepas_gateway/dbus/sink_manager.py
+++ b/python_transport/wirepas_gateway/dbus/sink_manager.py
@@ -499,6 +499,35 @@ class Sink:
 
         return ret
 
+    def clear_local_scratchpad(self):
+        ret = wmm.GatewayResultCode.GW_RES_OK
+        restart = False
+        try:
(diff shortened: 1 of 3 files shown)


teemupiiroinenwirepas:
Is there some use case when the scratchpad should be cleared e.g. should backend support it?

sgunes-wirepas:
> Is there some use case when the scratchpad should be cleared e.g. should backend support it?

Not that I know of, this was only needed to make internal testing more flexible.

Before this change, making an update request with no bytes would result in an error response. Might be good to check if there are any assumptions on the backend side related to this.

GwendalRaoul (changes requested):
Except the comment on the seq, I'm fine with the change

MauriKuorilehto (approved):

GwendalRaoul (approved):

PR #281, opened 2024-11-22
Implementation of the Keep Alive thread in the transport service.

LePailleurThibault:


sgunes-wirepas:
Maybe a more general question, but why is the timezone offset needed in the keepalive messages? In locations with daylight saving time, the offset from UTC would change throughout the year.

python_transport/wirepas_gateway/keep_alive_service/keep_alive_service.py
```diff
@@ -0,0 +1,245 @@
...
+
+        self.disconnected_sinks = set()  # All sinks that are detected as disconnected to the gateway
+        self.connected_sinks = set()  #  All sinks that are detected as connected to the gateway
+
+        self.keep_alive_interval_s = keep_alive_interval_s
+        self.keep_alive_timezone_offset_mn = keep_alive_timezone_offset_mn
+
+    def prepare_keep_alive_msg(self):
+        """ Prepare and return a keep alive message. """
+        rtc_timestamp_ms = None
```
GwendalRaoul, line 139: Are those lines necessary, they are updated unconditionally just after.

python_transport/wirepas_gateway/keep_alive_service/keep_alive_service.py
```diff
@@ -0,0 +1,245 @@
...
+
+import wirepas_mesh_messaging as wmm
+
+
+KEEP_ALIVE_SERVICE_VERSION = 0x01
+KEEP_ALIVE_SRC_EP = 67
+KEEP_ALIVE_DST_EP = 67
+
+# Timeouts and periods used for the keep alive service.
+WM_MSG_RETRY_PERIOD_S = 1
```
GwendalRaoul, line 19: Is 1s enough ?

python_transport/wirepas_gateway/keep_alive_service/keep_alive_service.py
```diff
@@ -0,0 +1,245 @@
...
+
+    def send_keep_alive_msg_to_sink(self, sink) -> bool:
+        """
+        Send the keep alive message to the network.
+        Returns True if the keep alive message could be sent to the sink, False otherwise.
+
+        Args:
+            sink: Sink to send the keep alive message to.
+        """
+        retries_left = 3
```
GwendalRaoul, line 162: As delay betwwen retries is a constant, why not the number of retires?

python_transport/wirepas_gateway/keep_alive_service/keep_alive_service.py
```diff
@@ -0,0 +1,245 @@
...
+            payload = keep_alive_message.encode_tlv()
+            logging.debug("Send the following keep alive payload to sink %s: %s",
+                          sink.sink_id, payload.hex())
+
+            res = sink.send_data(
+                dst=BROADCAST_ADDRESS,
+                src_ep=KEEP_ALIVE_SRC_EP,
+                dst_ep=KEEP_ALIVE_DST_EP,
+                qos=0,
+                initial_time=0,
```
GwendalRaoul, line 177: If I recall right, it is already the default value, so could be skipped

python_transport/wirepas_gateway/keep_alive_service/keep_alive_service.py
```diff
@@ -0,0 +1,245 @@
...
+        if start_timer:
+            time_to_wait = max(time_to_wait - (time() - start_timer), 0)
+
+        sleep(time_to_wait)
+
+    def run(self):
+        """ Main loop that send periodically keep alive message to the network. """
+        while True:
+            # Put a timer so that the message are periodic with a good precision
+            start_timer = time()
```
GwendalRaoul, line 203: Shouldn't time.monotonic() be used instead, to avoid issue if clock is changed during the process?

python_transport/wirepas_gateway/keep_alive_service/keep_alive_service.py
```diff
@@ -0,0 +1,245 @@
...
+            # Send keep alive messages to all sinks
+            logging.info("Send a keep alive message to the network")
+            for sink_id in current_sinks:
+                if self.send_keep_alive_msg_to_sink(self.sink_manager.get_sink(sink_id)):
+                    current_connected_sinks.add(sink_id)
+                else:
+                    self.disconnected_sinks.add(sink_id)
+
+            # Find sinks that disappeared and those which reconnected during this iteration
+            non_detected_sinks = self.connected_sinks.difference(current_sinks)
```
GwendalRaoul, line 223: What is the purpose of such mechanism?

python_transport/wirepas_gateway/keep_alive_service/keep_alive_service.py
sgunes-wirepas: Good point, monotonic would be much safer. How about the nodes, can it be a problem there if the keepalive message has an older timestamp than the previous one?

python_transport/wirepas_gateway/keep_alive_service/keep_alive_service.py
LePailleurThibault: Well if a message can't be sent to a sink 3 times in a row in ~2s, there are definitely issue with this sink. So not sending a keep alive message for that sink wouldn't be that bad.

python_transport/wirepas_gateway/keep_alive_service/keep_alive_service.py
LePailleurThibault: Yes, I will use a monotonic timer here. The gateway is supposed to be connected to a NTP server when running the keep alive service and the keep alive messages contain the UTC timestamp. This issue will never happen with this timestamp. Now if the application wants to use local time (based on UTC + timezone offset), the application itself should handle the consequence of DST.

sgunes-wirepas: I was thinking more about the following (very unlikely) scenario:
- NTP server(s) becomes unreachable during runtime, but not MQTT
- Host clock drifts forward
- NTP is reconnected and clock is updated
- New keepalive message has an older timestamp than the previous one

LePailleurThibault: Time should not be included in the keep alive packet if the NTP server time is unavailable. However, at this moment, we do not have any NTP service-like API for this kind of issue. So there is no way of knowing if this time is valid without implementing a NTP server nearby.
But at the moment, when using the time of this service, the gateway is supposed to be synchronized with a NTP server.
And if it the issue to happen, the time will not be updated in the current applications if the clock difference is too big between 2 packets.

python_transport/wirepas_gateway/keep_alive_service/keep_alive_service.py
```diff
@@ -0,0 +1,245 @@
...
+    def __init__(self, version, gateway_status=None, rtc_timestamp_ms=None,
+                 timezone_offset_mn=None, keep_alive_interval_s=None):
+        self.version = version
+        self.gateway_status = gateway_status
+        self.rtc_timestamp_ms = rtc_timestamp_ms
+        self.timezone_offset_mn = timezone_offset_mn
+        self.keep_alive_interval_s = keep_alive_interval_s
+
+    @classmethod
+    def encode_tlv_item(cls, elt_type, length, value, packing):
```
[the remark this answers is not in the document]
GwendalRaoul: @staticmethod
Is probably the best choice IMO for this purpose. You don't use the cls first paraeter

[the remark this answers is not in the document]
LePailleurThibault: Good point!

python_transport/wirepas_gateway/keep_alive_service/keep_alive_service.py
LePailleurThibault: This argument is required by the method. I don't think there is any default value for the initial_time.

LePailleurThibault:
This PR is outdated.
The keep alive service is now part of the amiv2 releases.
See amiv2_vx.y.z tags

PR #283, opened 2024-12-31
Update docker image and build process for transport service

sgunes-wirepas:
Changes are mainly aimed at updating dependencies and to make local building easier. Protobuf python dependency is built in place with the UPB backend, so we don't need to use the base image from [wirepas-mesh-messaging-python](https://github.com/wirepas/wirepas-mesh-messaging-python).

Please see individual commit messages for more details.


GwendalRaoul (approved):
Very good improvement!

PR #292, opened 2025-01-31
Transport: Simplify setStatus mechanism

GwendalRaoul:
It should avoid race condition to update status
And it also introduce a backup mechanism to ensure a check of config every hour

Closes # .

### DIFF
--- a/python_transport/wirepas_gateway/transport_service.py
+++ b/python_transport/wirepas_gateway/transport_service.py
@@ -8,7 +8,7 @@ import sys
 import wirepas_mesh_messaging as wmm
 from time import time, sleep
 from uuid import getnode
-from threading import Lock, Thread
+from threading import Thread, Event
 
 from wirepas_gateway.dbus.dbus_client import BusClient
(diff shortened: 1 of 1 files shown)


python_transport/wirepas_gateway/transport_service.py
```diff
@@ -181,6 +181,139 @@ def initialize_sink(self, name):
...
+                # Reset attempts counter
+                attempt = 0
+            else:
+                logging.info("Backup mechanism to send a new status or next attempt")
+
+            # Before generating config, reset the Event to be sure any
+            # request will be served after this point
+            self._set_status_event.clear()
+
+            if self._set_status() and attempt < 5:
```
sgunes-wirepas, line 287: 5 is used multiple times, can be a local variable to avoid errors if it is modified

sgunes-wirepas (approved):

python_transport/wirepas_gateway/transport_service.py
GwendalRaoul: Of course !

PR #296, opened 2025-02-05
Add scanning neighbors feature

kowalewskijan:
This PR adds a possibility to initiate scanning over Dbus and to get scanning results back.

Scope of changes:

- Changes to sink_service to enable network scanning feature over dbus
- Changes to sink_manager Python layer to provide scanning network API


kowalewskijan:
@GwendalRaoul @sgunes-wirepas Could you PTAL? It's a relatively small change

GwendalRaoul:
Hello @kowalewskijan,

Sorry for the delay to answer but we had some internal discussion first.

Even if your change is correct, it is not something we want to promote in our sinkService. In your case it may be working fine, but in general, forcing scan on a sink may result in performances degradation as the scan will prevent routing and flushing messages from the network.
Furthermore, the expectation on scan length is not valid for all profiles. It should be configurable.
And the list of neighbor generated by the stack, is based on stack own criteria that are not always the one expected by users (like best RSSI or best link quality).

So for now we will not merge this PR in our master. I hope you will understand it.

Best regards,
Gwendal

kowalewskijan:
Hello @GwendalRaoul,

What is then the best way to discover nodes attached to the sink in order to send data to them?

BR,
Jan

kowalewskijan:
@GwendalRaoul feel free to close this PR if it's `won't merge`. But I'm wondering what's the purpose of neighbors discovery functionality if it's that troublesome?

GwendalRaoul:
Hello,

What would be the use case associated to it?

Are you looking for nodes attached "directly" to sinks or belonging to a sink tree?
In a mesh network, devices may report through a sink but be very far (3,4,5... x radio range apart). So gateway or backend should maintain a list of devices with their last reporting sink.
This way if you want to send downlink traffic you know from which sink to send.

The api you are using returns a partial list of devices in same radio range. But it is not exhaustive list and some nodes mays even not be reporting to the sink directly.

Best regards,
Gwendal

kowalewskijan:
Hello @GwendalRaoul,

I have a FOTA service which sends packets over data plane to 3rd party chip connected through UART to nRF52 with dualmcu application. The service is run from the Gateway which has a sink connected, so I'd have to know if the node which supports this FOTA service is reachable. Of course the service could just time out if it's not reachable, but the first approach was to use this scanning network feature to see if the node is associated with the sink from which the service will be run.

BR,
Jan

PR #305, opened 2025-08-20
Support for sink provisioning

sgunes-wirepas:
Adds support for setting network and management keys using the new API on  the sink.

Additionally:

- Hide keys from transport service logs.
- Update build system for sink service. CMake is used, and c-mesh-api dependency is downloaded via cmake instead of being a git submodule. Earlier, sink service was built without optimization flags. Now -O2 is used.


docker/sink_service/Dockerfile
```diff
@@ -6,7 +6,7 @@ ARG GATEWAY_BUILD_SHA1=unset
 
 RUN adduser --disabled-password wirepas
 
-RUN apk add --no-cache gcc make cmake musl-dev dpkg elogind-dev linux-headers
+RUN apk add --no-cache gcc git make cmake musl-dev dpkg elogind-dev linux-headers
```
MauriKuorilehto, line 9: Git required now here in the builder due to the change related to submodule usage?

python_transport/wirepas_gateway/transport_service.py
```diff
@@ -365,6 +366,36 @@ class TransportService(BusClient):
...
+        """
+        Wrapper class to show only allowed configuration fields of
+        SetConfigRequest. Can be used to exclude potentially sensitive fields
+        from logs.
+        """
+        def __init__(self, request: wmm.SetConfigRequest):
+            self.orig_request = request
+
+        def __str__(self):
+            allowed_fields = [
```
MauriKuorilehto, line 379: Could be also other way around, i.e. `masked_fields`, so that possible new fields does not require change in here? I guess anyways most of the fields will be allowed.

sink_service/CMakeLists.txt
```diff
@@ -0,0 +1,33 @@
...
+endif()
+set(CMAKE_C_FLAGS_RELEASE "-O2")
+
+add_compile_options(-Wall -Werror -Wextra -Wno-unused-parameter)
+
+include(FetchContent)
+FetchContent_Declare(
+    c-mesh-api
+    GIT_REPOSITORY https://github.com/wirepas/c-mesh-api/
+    GIT_TAG c566fdbd693eb3fb6027286679f53db9f9958e0a
```
MauriKuorilehto, line 17: I wonder whether this should be extracted from here to some "config" file? Also should we use some explicit tags instead of commit hashes?

sink_service/source/config.c
```diff
@@ -776,6 +776,129 @@ static int get_config_data_content(sd_bus_message *m, void *userdata, sd_bus_err
...
+        const app_res_e wpc_res = APP_RES_INVALID_VALUE;
+        SET_WPC_ERROR(error, __FUNCTION__, wpc_res);
+        LOGE("Invalid authentication key size (%zu) (ret=%d)\n", key_size, wpc_res);
+        return -EINVAL;
+    }
+
+    memcpy((void*)security_keys->key_pair.authentication, key_bytes, sizeof(security_keys->key_pair.authentication));
+
+    r = sd_bus_message_read(message, "y", &security_keys->sequence_number);
+    if (r < 0)
```
MauriKuorilehto, line 836: Should we also here validate that the sequence falls into the supported range `[1..255]`? Or does we let sink node itself to handle that (or will the APIs just silently crop the value)?

python_transport/wirepas_gateway/transport_service.py
sgunes-wirepas: It can be, and it would even be easier and simpler to do. But I thought this is safer in case a new field is added in the future and we forget to add it to the filter to hide.

sink_service/CMakeLists.txt
sgunes-wirepas: I think tags would be good too. There are some tags/releases for c-mesh-api from the past(2022) which were using the same version as gateway. I don't know if there would be an advantage to use the same version numbers, but we can start creating releases for c-mesh-api again.

python_transport/wirepas_gateway/transport_service.py
MauriKuorilehto: Yeap, in theory I agree, in practice it then means that we need to always add fields to this (which may not be that obvious). I guess there should not be that many secrets that we pass through this API that I would maybe still lean towards turning this around.

sgunes-wirepas: Alright I can change it, I agree probably there won't be such fields coming anytime soon anyway.

PR #306, opened 2025-09-05
docker compose example with included MQTT broker

sgunes-wirepas:


docker/docker-compose/single_transport_local_mqtt/README.md
```diff
@@ -0,0 +1,11 @@
+# Single transport with local MQTT broker example
+
+This example is based on the [single transport example](../single_transport/README.md).
+The configuration instructions from single transport can be followed. In this
+example, the MQTT broker is added to the docker compose file as a container, so
+you don't have to set up and configure an MQTT broker separately.
+
+The MQTT broker is configured insecurely, therefore this example is mainly
```
GwendalRaoul, line 8: And it should be stipulated also that it is a very particular setup.

In fact, it only allows a single gateway in a network. It is good for prototyping but not a real use case.

GwendalRaoul (changes requested):

PR #308, opened 2025-09-17
Setting sink service log level using environment variables

sgunes-wirepas:
Log levels for individual modules in sink service can now be set using environment variables.

### DIFF
--- a/docker/docker-compose/README.md
+++ b/docker/docker-compose/README.md
@@ -89,3 +89,49 @@ transport-service:
       WM_SERVICES_MQTT_HOSTNAME: "aBroker"
 ```
 
+## Sink service
+Similar to transport service, sink service can also be configured with
+environment variables. Below is a list of parameters for the sink service:
+
+
+| Parameter                          | Description                                                                                     | Default value           | Possible value       |
+| ---                                | ---                                                                                             | ---                     | ---                  |
+| WM_GW_SINK_BAUDRATE                | UART BAUD rate for sink. Detected automatically from default values if not provided.            | 125000, 115200, 1000000 | non-negative integer |
+| WM_GW_SINK_BITRATE                 | Alternative alias for WM_GW_SINK_BAUDRATE                                                       | 125000, 115200, 1000000 | non-negative integer |
+| WM_GW_SINK_ID                      | Sink ID                                                                                         | 0                       | non-negative integer |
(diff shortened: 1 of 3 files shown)


docker/docker-compose/README.md
```diff
@@ -89,3 +89,49 @@ transport-service:
...
+| WM_GW_SINK_MAX_FRAGMENT_DURATION_S | Maximum duration in seconds to keep fragment from incomplete data packets. Zero equals forever. | 900                     | non-negative integer |
+| WM_GW_SINK_DOWNLINK_LIMIT          | Max number of downlink messages being queued in parallel. Zero equals no limit.                 | 0                       | 0-16                 |
+| WM_DEBUG_LEVEL                     | Global log level. See section "Setting log level" for more information.                         | INFO                    | string               |
+| WM_MODULE_DEBUG_LEVEL              | Module specific log levels. See section "Setting log level" for more information.               | -                       | string               |
+
+### Setting log level
+The log level can be set with the WM_DEBUG_LEVEL and WM_MODULE_DEBUG_LEVEL environment variables.
+
+The following log levels are possible:
+ - QUIET
```
GwendalRaoul, line 113: any reason to use quiet instead of No_log ?
I'm fine with both, but in the conversion, this is the only one that is different :-)

docker/docker-compose/README.md
```diff
@@ -89,3 +89,49 @@ transport-service:
...
+
+To have a reasonable amount of debug logs, the parameters can be set to the
+following values:
+```bash
+WM_DEBUG_LEVEL="DEBUG"
+WM_MODULE_DEBUG_LEVEL="SLIP:INFO;linux_plat:INFO;wpc_int:INFO"
+```
+This would set the log level of every module to "DEBUG", and then set the log
+level of modules "SLIP", "linux_plat", and "wpc_int" to "INFO".
+
```
GwendalRaoul, line 137: Maybe could be good to add a note that evaluation of global variable is only done at starting of service.
It cannot be changed on the fly 
Or do we want to add it by reading again the global variable every 30s for example? That could be very handy because when investigating a bug, sometimes a restart suppress it.
Ideally it would be good to be notified when env has been updated, but not aware of such service in Linux ;-)

GwendalRaoul: Looks like it could be hard to achieve and not sure that it is straight forward to modify an env var in a running docker container.

GwendalRaoul (approved):

PR #311, opened 2025-11-10
LG-91 Stop stack before setting security keys on sink

sgunes-wirepas:


python_transport/wirepas_gateway/dbus/sink_manager.py
```diff
@@ -248,6 +248,14 @@ def _set_param(self, dic, key, attribute):
 
     def _set_network_keys(self, config):
         if "network_keys" in config:
+            # Stop the stack if not already stopped
+            try:
+                if self.proxy.StackStatus == 0:
+                    self.proxy.SetStackState(False)
```
MauriKuorilehto: Where do we restart the stack?

PR #317, opened 2026-03-12
Update paho-mqtt to v2.1.0

sgunes-wirepas:
Using library callback API version 1 which is deprecated for an easier transition. Also explicitly setting the protocol version to v3.1.1 (which is the default anyway) because current callbacks won't work with MQTTv5.

Removing the _sockpairR workaround because in paho-mqtt v2.1.0, it is not initialized when not using client.loop() or client.loop_start(), which is our case and therefore not needed anymore.

mqtt_wrapper.start() is moved to the end of the constructor of TransportService class because if MQTT CONNACK is received too soon, transport service was crashing due to status_thread not being initialized.

Also fixing the callback function signature for _on_disconnect.

Note: _sockpairR issue can be tested by running the transport service and investigating the local tcp sockets with `ss -tmp | grep wm-gw`. With the older version if you disable the workaround, send buffer of one of the sockets fills up as messages are published. With the new version, the extra 2 sockets don't get created.

### DIFF
--- a/python_transport/wirepas_gateway/protocol/mqtt_wrapper.py
+++ b/python_transport/wirepas_gateway/protocol/mqtt_wrapper.py
@@ -52,6 +52,8 @@ class MQTTWrapper(Thread):
             self._use_websockets = False
 
         self._client = mqtt.Client(
+            callback_api_version=mqtt.CallbackAPIVersion.VERSION1,
+            protocol=mqtt.MQTTProtocolVersion.MQTTv311,
             client_id=settings.gateway_id,
             clean_session=not settings.mqtt_persist_session,
             transport=transport,
@@ -133,7 +135,7 @@ class MQTTWrapper(Thread):
         if self.on_connect_cb is not None:
             self.on_connect_cb()
 
-    def _on_disconnect(self, userdata, rc):
+    def _on_disconnect(self, client, userdata, rc):
         if rc != 0:
             logging.error(
                 "MQTT unexpected disconnection (network or broker originated):"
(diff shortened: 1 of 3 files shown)


python_transport/wirepas_gateway/protocol/mqtt_wrapper.py
sgunes-wirepas, line 149: Also good to point out that with the new version, exceptions thrown in the callbacks are raised again. So any uncaught exception in a callback would crash the connection. In this case the "mid" is always added to the set before a publish, so there shouldn't™ be a key error here when removing from the set.

python_transport/wirepas_gateway/protocol/mqtt_wrapper.py
```diff
@@ -52,6 +52,8 @@ def __init__(
             self._use_websockets = False
 
         self._client = mqtt.Client(
+            callback_api_version=mqtt.CallbackAPIVersion.VERSION1,
+            protocol=mqtt.MQTTProtocolVersion.MQTTv311,
```
jussijii1, line 56: Should we have a statement in the top level doc., and in an other relevant service doc. that is using MQTT implementation, that the we use fixed version MQTT 3?

python_transport/wirepas_gateway/transport_service.py
```diff
@@ -481,6 +477,10 @@ def __init__(self, settings, **kwargs):
         # Dictionnary to store scratchpad chunks
         self._scratchpad_chunks = {}
 
+        self.mqtt_wrapper.start()
```
jussijii1, line 480: "mqtt_wrapper.start() is moved to the end of the constructor of
TransportService class because if MQTT CONNACK is received too soon,
transport service was crashing due to status_thread not being
initialized."
If this is true, that self.mqtt_wrapper.start() is moved to end of the ctor due to the fact that other async operation needs to be finished until self.mqtt_wrapper.start() is called, we probably should have some event based system. Synchronizing two async clock domains with timing based system can lead potential issues.
Or is it moved end of the ctor due to self.status_thread.start() call?
Anyways, this comment is a suggestion and might be too big of a change for this commit.

jussijii1:
General comment, some of the functions are documented and some are not. We probably should start documenting the functions, especially in those functions where the changes are done, to keep the style consistent.

jussijii1 (approved):
I left couple of comments, mostly about general improvements, otherwise the PR looks good for me.

python_transport/wirepas_gateway/transport_service.py
sgunes-wirepas: The problem was that status_thread was not existing, so not necessarily the timing of status_thread.start() call which was the issue. At least for the mqtt and status thread connection things should be okay I think.

However, there might be some timing issues with dbus initialization. We can get a signal for on_stack_started or on_stack_stopped before the status thread is initialized and those need the status thread. Parent BusClient constructor is called in the beginning via super(), and status_thread is created later.

Might be worth testing at some point by starting the transport service while spamming the dbus signals (on data received, on stack stopped, on stack started).

PR #318, opened 2026-03-13
Fix build for rtc_service and local_history_service

sgunes-wirepas:
Transport service wheel is used in these services for its dbus client (as mentioned in the comment in the dockerfiles). With recent changes, transport service build needs to be given the version.

A dummy version is now given when building the transport service wheel for these services. This is okay because the version of the transport service is not used in these services.

Should fix the following failed builds:
https://github.com/wirepas/gateway/actions/runs/23045402437
https://github.com/wirepas/gateway/actions/runs/23045402494

### DIFF
--- a/docker/local_history_service/Dockerfile
+++ b/docker/local_history_service/Dockerfile
@@ -23,7 +23,7 @@ WORKDIR /home/wirepas
 COPY --chown=wirepas ./python_transport /home/wirepas/python_transport
 WORKDIR /home/wirepas/python_transport
 
-RUN ./utils/generate_wheel.sh
+RUN SETUPTOOLS_SCM_PRETEND_VERSION_FOR_WIREPAS_GATEWAY=0.0.0 ./utils/generate_wheel.sh
 
 USER wirepas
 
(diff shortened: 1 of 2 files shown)


docker/local_history_service/Dockerfile
```diff
@@ -23,7 +23,7 @@ WORKDIR /home/wirepas
 COPY --chown=wirepas ./python_transport /home/wirepas/python_transport
 WORKDIR /home/wirepas/python_transport
 
-RUN ./utils/generate_wheel.sh
+RUN SETUPTOOLS_SCM_PRETEND_VERSION_FOR_WIREPAS_GATEWAY=0.0.0 ./utils/generate_wheel.sh
```
jussijii1, line 26: Just thinking out loud, could be there be internal default value, which generate_wheel.sh uses, instead of giving explicit default value from the command line and to override the default value in case when it's needed?

sgunes-wirepas: Yeah it's possible. The reason I didn't add it is to prevent having a dummy version by accident when we actually need a version.

Normally if you run that script on a cloned repository locally without the environment variable, the build plugin will figure out a version based on git tags. Here it fails because we copy a subdirectory and there isn't any git info.

jussijii1: Ok, thanks for the information.

jussijii1 (approved):
Looks fine for me.

PR #321, opened 2026-03-27
Print commit hash as sink service version

sgunes-wirepas:
The sink service and c-mesh-api commit hashes are now printed during sink service startup. This makes it easier to check what version has been running when using the "edge" tagged docker images.

Note: Earlier docker images also had this information in the label of the image, which can be checked with:
`docker inspect --format '{{.Config.Labels}}' wirepas/gateway_sink_service:edge`


jussijii1 (approved):
Looks good to me

PR #323, opened 2026-05-07
Set cost for existing sinks also when black hole detection is disabled

sgunes-wirepas:
Existing sinks are detected at startup in the constructor of SinkManager by listing bus objects matching the sink service names.

Earlier, the cost for these existing sinks was only initialized when the black hole detection was enabled, when ConnectionToBackendMonitorThread is run.

After the sink cost has been increased due to MQTT disconnection, if the gateway is stopped before MQTT is re-connected, we end up with a sink with increased cost. At this point if the gateway is reconfigured to disable the black hole detection sink can get stuck with the increased cost.

To avoid that, the sink costs are now initialized also if the ConnectionToBackendMonitorThread is not started.

### DIFF
--- a/python_transport/wirepas_gateway/dbus/sink_manager.py
+++ b/python_transport/wirepas_gateway/dbus/sink_manager.py
@@ -418,6 +418,8 @@ class Sink:
         if new_cost is None or new_cost < 0 or new_cost > 254:
             raise ValueError("Wrong sink cost value {}".format(new_cost))
 
+        logging.debug(f"Setting cost {new_cost} for sink {self.sink_id}")
+
         try:
             self.proxy.SinkCost = new_cost
         except GLib.Error as err:
@@ -842,3 +844,8 @@ class SinkManager:
         except KeyError:
             logging.error("Unknown sink %s from sink list", short_name)
             return None
+
+    def set_sink_costs(self, cost):
+        for sink in self.get_sinks():
+            sink.cost = cost
+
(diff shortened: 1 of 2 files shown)


python_transport/wirepas_gateway/dbus/sink_manager.py
```diff
@@ -842,3 +844,8 @@ def get_sink(self, short_name):
         except KeyError:
             logging.error("Unknown sink %s from sink list", short_name)
             return None
+
+    def set_sink_costs(self, cost):
```
jussijii1, line 848: This is setting sink cost for all the sinks in one batch. Do we have or need functionality to set sink cost for a given sink?

jussijii1, line 848: The code is self explanatory but it would be good to document the code at least in function level. sets all the sinks for the given sink cost. Documentation in general could be improved in that sense.

jussijii1 (approved):
A comment about general function level documentation and sink cost per sink ID.
The change as such is OK but I'm thinking, can there be potential issue in some scenario where multiple sinks are used by the gateway and sink cost is set for all the sinks in one batch.

python_transport/wirepas_gateway/dbus/sink_manager.py
sgunes-wirepas: In the code yes we can set cost for a single sink. It is used when a new sink service is started after transport service, in initialize_sink() in transport_service.py. Otherwise at startup and on MQTT connect/disconnect, all the existing sinks cost is being set.

sgunes-wirepas: Added documentation for this one only. We can add missing documentations for edited and new functions going forward.

GwendalRaoul (approved):

PR #325, opened 2026-05-29
LG-99 Only warn when sink cost cannot be set due to invalid role

sgunes-wirepas:
Starting with commit 1a3eae3 sink cost is set for existing sinks at startup. This can fail if the node role is not sink, for example with a freshly flashed node.

Before this fix, transport service would crash at startup if the node was not a sink because of this. Now these failures are only warnings and the transport service can be started with a node with non-sink role.

### DIFF
--- a/python_transport/wirepas_gateway/transport_service.py
+++ b/python_transport/wirepas_gateway/transport_service.py
@@ -656,10 +656,7 @@ class TransportService(BusClient):
                         name, self.minimum_sink_cost
                     )
                 )
-                try:
-                    sink.cost = self.minimum_sink_cost
-                except ValueError:
-                    logging.debug("Cannot set cost, probably not a sink")
+                sink.cost = self.minimum_sink_cost
 
     @update_gateway_status_dec
     def on_sink_disconnected(self, name):
--- a/python_transport/wirepas_gateway/dbus/sink_manager.py
+++ b/python_transport/wirepas_gateway/dbus/sink_manager.py
@@ -426,7 +426,6 @@ class Sink:
             res = ReturnCode.error_from_dbus_exception(err.message)
             if res == wmm.GatewayResultCode.GW_RES_INVALID_ROLE:
                 logging.warning("Node role is not a sink, sink cost cannot be modified")
-                raise ValueError("Wrong role to set cost value {}".format(new_cost))
             else:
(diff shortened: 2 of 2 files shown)


python_transport/wirepas_gateway/transport_service.py
```diff
@@ -656,10 +656,7 @@ def on_sink_connected(self, name):
                         name, self.minimum_sink_cost
                     )
                 )
-                try:
```
jussijii1, line 659: Just thinking out loud one thing, if we here are unsure that the instance might not be really a sink, even though that it is fetched by sink manager, do we have at sink manager logic to ensure that the node is for sure a sink.

GwendalRaoul (approved):

jussijii1 (approved):
"sink cost is set for existing sinks at startup. This can fail if the node role is not sink, for example with a freshly flashed node."
Is there a way in GW to fetch node information to ensure that the node is sink before doing any actions for it? The logic in the commit comment little bit of hint that perhaps we miss something. We set sink cost for a sink but it might not be a sink.

sgunes-wirepas:
@jussijii1 I think terminology makes it more confusing. For gateway, when we say "sink", it is rather a node implementing the dualmcu api. It is expected to have a sink role for the mesh+gateway to function, but the interface allows to have a node with non-sink role. Since the role can be changed any time, checking for it at startup might not help.

Maybe a bigger topic is that is it even a valid use case to have a "sink" with non-sink role connected to the gateway? I suppose changing the role to non-sink role can have a similar effect as raising the cost, therefore can be a valid? But I don't think we have a definition on how the gateway should behave in that case, because it can't read every parameter.

jussijii1:
> @jussijii1 I think terminology makes it more confusing. For gateway, when we say "sink", it is rather a node implementing the dualmcu api. It is expected to have a sink role for the mesh+gateway to function, but the interface allows to have a node with non-sink role. Since the role can be changed any time, checking for it at startup might not help.
> 
> Maybe a bigger topic is that is it even a valid use case to have a "sink" with non-sink role connected to the gateway? I suppose changing the role to non-sink role can have a similar effect as raising the cost, therefore can be a valid? But I don't think we have a definition on how the gateway should behave in that case, because it can't read every parameter.

Thanks for the input.

PR #327, opened 2026-07-09
Limit size of SelectableQueue in transport service

sgunes-wirepas:
See individual commit messages for more details.

Briefly:
- Deprecated WM_GW_BUFFERING_STOP_STACK parameter, it is replaced by WM_GW_BUFFERING_ACTION
- Added new buffering action to drop packets. If MQTT connection is lost, the sink is not modified and the size of the internal queue is limited.

### DIFF
--- a/python_transport/wirepas_gateway/transport_service.py
+++ b/python_transport/wirepas_gateway/transport_service.py
@@ -15,6 +15,7 @@ from wirepas_gateway.dbus.dbus_client import BusClient
 from wirepas_gateway.protocol.topic_helper import TopicGenerator, TopicParser
 from wirepas_gateway.protocol.mqtt_wrapper import MQTTWrapper
 from wirepas_gateway.utils import ParserHelper
+from wirepas_gateway.utils.argument_tools import BufferingAction
 
 from wirepas_gateway import __version__ as transport_version
 from wirepas_gateway import __pkg_name__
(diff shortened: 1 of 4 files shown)


python_transport/wirepas_gateway/transport_service.py
```diff
@@ -163,8 +176,8 @@ def initialize_sink(self, name):
         Args:
             name: name of sink to initialize
         """
-        # It is only required if black hole is managed by sink cost
-        if not self.stop_stack:
+        # It is only required if black hole is managed by sink cost.
+        if self.buffering_action == BufferingAction.RAISE_SINK_COST:
```
mrceyhun: There is no `buffering_action` in `ConnectionToBackendMonitorThread` class, but it has `stop_stack` https://github.com/wirepas/gateway/pull/327/changes#diff-3b9500134876e39c4c65502cfb4f555155db6858a64aea507bbe617fca648415L39

mrceyhun (changes requested):
I did a quick review and it looks fine for me if we also add new parameters explanations in https://github.com/wirepas/gateway/blob/master/docker/docker-compose/README.md

python_transport/wirepas_gateway/transport_service.py
sgunes-wirepas: Good find, thanks! Initially I passed buffering_action to that class, but then decided to keep it untouched - forgot that part.

sgunes-wirepas: Thanks for checking, updated the readme now. I think for the future it would be good to update the documentation to have feature level information instead of just parameter level, at least for this buffering feature it would be needed. Probably would be good to move the documentation somewhere else as well, but we can discuss that later :)

mrceyhun: Thanks @sgunes-wirepas

mrceyhun (approved):

python_transport/wirepas_gateway/transport_service.py
```diff
@@ -1231,6 +1245,20 @@ def _update_parameters(settings):
             logging.error("Wrong format for whitened_endpoints_filter EP list (%s)", e)
             exit()
 
+    if settings.buffering_stop_stack is not None:
+        logging.warning("Param buffering_stop_stack is deprecated, please use buffering_action instead")
+        if settings.buffering_action is not None:
+            logging.error("Param buffering_stop_stack and buffering_action cannot be set at the same time")
```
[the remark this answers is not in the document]
jussijii1: Perhaps little bit of matter of taste, but I'd prefer if on line 1250 with same indentation as if on line 1248 and have 
`if settings.buffering_stop_stack is not None and settings.buffering_action is not None`
For me it improves readability since the case when the error is given is more explicit.

jussijii1 (approved):
Added one minor comment considering syntax-semantic sugar. Thanks to let me know about the change

